PR #212, opened 2014-10-15
Fixes test on Windows OS

q42jaap:
Because of git checkout, the source code with """ multi-line string literals changes meaning :(

Also, the `.toList.flatten` construct was not understood by my IDE (intellij). I changed it to
`.getOrElse(List.empty)` because imho that is the nicest way to convert `Option[List[Album]]`
to `List[Album]`. Made intellij happy too :)


sgodbillon:
Thanks!

PR #229, opened 2014-12-17
Add `BSONCodec[A]` alias.

aloiscochard:


bson/src/main/scala/bson.scala
```diff
@@ -17,6 +17,7 @@ package reactivemongo.bson
 
 object `package` extends DefaultBSONHandlers {
   type BSONElement = (String, BSONValue)
+  type BSONCodec[A] = BSONDocumentReader[A] with BSONDocumentWriter[A] with BSONHandler[BSONDocument, A]
```
sgodbillon: I think `BSONDocumentHandler` is more consistent with `BSONHandler`.  @aloiscochard What do you think?

aloiscochard: @sgodbillon agree! I just used that name because I saw it other lib, but indeed what you suggest is more consistent with `ReactiveMongo`

jbgi:
looks like this has been fixed via https://github.com/ReactiveMongo/ReactiveMongo/pull/612/commits/1a76adf929b08c3d94a0219fe1d1ccce89e47c82

PR #230, opened 2014-12-17
Feature/handlers

aloiscochard:
Miscellaneous handlers improvements.


aloiscochard:
@sgodbillon It's terrible man, there is an horrible bug here that I have not catched until now... I feel shameful.

That explain why as well I had the feeling that reactivemongo was bloking sometime... it's hard to catch because it's a shadowing issue here:
https://github.com/aloiscochard/ReactiveMongo/commit/88f918fab7c117fe31715e76311c0b5cef9ed3ce#diff-80a726893b72e4ce844857db08243368R104

Which make the method go in infinite recursion...

Pardon me, I hope you did not cut a release with this?

The fix is trivial, I'll prepare a PR with it and a few other additions.

I'm sorry :-(

PR #238, opened 2014-12-23
Fix Equality And Other Issues

reid-spencer:
This patch adds a test suite for testing equality of BSON classes including after serialization/deserialization. The idea is that the objects should be reflective after storage in the db and later materialized back again. Before this patch, this wasn't true. 

* Implement BSONDBPointerBufferHandler.write
* Make ArrayReadableBuffer a case class to get properly constructed .canEqual and .equals methods to avoid false negatives
* Fix misleading comments in BSONArray
* Fix misleading toString method in BSONArray (reported it was a BSONDocument)
* Make BSONObjectID extend Equals and implement canEqual and equals for it, properly (gave false negatives previously)
* Make BSONDBPointer override equals and canEqual to avoid false negatives
* Add an Equality test suite to validate that the above all works.

### DIFF
--- a/bson/src/main/scala/buffer.scala
+++ b/bson/src/main/scala/buffer.scala
@@ -153,7 +153,7 @@ import java.nio.ByteBuffer
 import java.nio.ByteOrder._
 
 /** An array-backed readable buffer. */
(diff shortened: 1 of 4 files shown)


sgodbillon:
Thanks!

PR #240, opened 2014-12-23
Minimize warnings

reid-spencer:
Reduce Compiler Warnings & Fix A Bug 
* Include the SBT modernization patch
* Don't cross compile, just use Scala 2.11.4
* akka-actor: 2.3.6 => 2.3.8
* play-iteratees: 2.3.5 => 2.3.7
* Fix the deprecation of Default.defaultSettings in build
* Introduce scala language options as needed to reduce warnings 
* Fix many deprecated Scala macro uses 
* Fix a bug in MongoDBSystem.postStop to avoid calling failure on an already completed 
  promise which previously threw IllegalStateException("Promise already completed")

### DIFF
--- a/bson/src/main/scala/bson.scala
+++ b/bson/src/main/scala/bson.scala
@@ -15,6 +15,8 @@
  */
 package reactivemongo.bson
 
+import scala.language.implicitConversions
+
 object `package` extends DefaultBSONHandlers {
   type BSONElement = (String, BSONValue)
 }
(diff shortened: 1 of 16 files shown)


sgodbillon:
Everything looks good to me.

Just wondering about 2.10 – 2.10.5 seems to be the last minor version in 2.10.x series.  I don't have any objection regarding dropping support for the next major release of ReactiveMongo.  Thoughts?

reid-spencer:
@sgodbillon - I think it is fine to drop support for 2.10.x in 0.11 release. Most people are on 2.11 these days and there are significant advantages to it. If they need to use 2.10.x, they can always use the 0.10 release.

PR #261, opened 2015-01-20
Fixed problem with indexes when ignoring field and added implementation to assign default value for ignored fields.

lucasrpb:


sgodbillon:
@lucasrpb  I'm sceptical about assigning a default value to ignored fields that way.  It seems to work only for primitive types, is that right? If so, this is a really narrow use case.

lucasrpb:
@sgodbillon I know ! But for another types (that are not primitive), it assigns "null". The only danger cases are the primitives like Boolean and Int. The rest of the classes are eligible to assign null, right? For example: BigInt. Unless the Scala developers create another primitive type :P

cchantep:
Thanks for the PR. Closing it for now. Let consider more feedback on the MailingList, to see whether it should be reopened. Best regards.

PR #270, opened 2015-02-04
Add `bsonCodecSeq`.

aloiscochard:
@sgodbillon I'm curious why there is no meta-codec defined that way?

Without it, I don't get codec for a `Seq[A]` if I have one for `A`.

I have the feeling that ` implicit def collectionToBSONArrayCollectionWriter` and `implicit def bsonArrayToCollectionReader` don't work in my case, due to the inheritance hierarchy (I have in fact defined a Handler, and those implicits don't work well with sub-typing.

I understand that you might refuse that PR in case it create ambiguity, but I wanted to show you the problem as that might affect a future redesign.

### DIFF
--- a/bson/src/main/scala/handlers.scala
+++ b/bson/src/main/scala/handlers.scala
@@ -196,6 +196,15 @@ trait DefaultBSONHandlers {
     new BSONArrayCollectionReader
   }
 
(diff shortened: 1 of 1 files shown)


cchantep:
Hi @aloiscochard. Thanks for that. I close it for now, not because it's refuse but better to consider/reopen when the BSON typeclasses will be refactored (hope for 0.12). Best regards.

PR #271, opened 2015-02-06
Avoid None.get when deleting an index.

oleastre:
Even if #130 adjusted the property name, I was still getting errors with mongo 2.6. 
Here is a patch against 0.10.5 that corrects the expected type of the nIndexesWas field.

### DIFF
--- a/driver/src/main/scala/core/commands/commands.scala
+++ b/driver/src/main/scala/core/commands/commands.scala
@@ -592,7 +592,7 @@ case class DeleteIndex(
 
   object ResultMaker extends BSONCommandResultMaker[Int] {
     def apply(document: BSONDocument) =
-      CommandError.checkOk(document, Some("deleteIndexes")).toLeft(document.getAs[BSONInteger]("nIndexesWas").map(_.value.toInt).get) }
+      CommandError.checkOk(document, Some("deleteIndexes")).toLeft(document.getAs[BSONNumberLike]("nIndexesWas").map(_.toInt).get) }
 }
 
 /** eval command */
--- a/driver/src/main/scala/core/commands/commands.scala
+++ b/driver/src/main/scala/core/commands/commands.scala
@@ -592,7 +592,7 @@ case class DeleteIndex(
 
   object ResultMaker extends BSONCommandResultMaker[Int] {
     def apply(document: BSONDocument) =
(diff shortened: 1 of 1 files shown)


driver/src/main/scala/core/commands/commands.scala
```diff
@@ -592,7 +592,7 @@ case class DeleteIndex(
 
   object ResultMaker extends BSONCommandResultMaker[Int] {
     def apply(document: BSONDocument) =
-      CommandError.checkOk(document, Some("deleteIndexes")).toLeft(document.getAs[BSONDouble]("nIndexesWas").map(_.value.toInt).get) }
+      CommandError.checkOk(document, Some("deleteIndexes")).toLeft(document.getAs[BSONInteger]("nIndexesWas").map(_.value.toInt).get) }
```
sgodbillon: It may be preferable to use `BSONNumberLike` instead, then converting it to `Int`.  Then it will work in any case.

oleastre:
Here is an updated version using BSONNumberLike...

cchantep:
Hi @oleastre , would you mind to rebase this one with recent changes on master?

oleastre:
This one targets the current stable branch 0.10.5.x.akka23 in the hope of a 0.10.5 minor release.
If an update to the 0.10.5 is not intended, I will rebase this PR on master. 
Otherwise I can create a new PR for the master branch (although from what I have seen, the api.DropIndexes command already handles this correctly).
What do you prefer ?

cchantep:
I guess this one can be rebase against master, as the command will stay in the next release, but as deprecated.

oleastre:
As a simple rebase included the release change and as the github interface did not let me change the branch this PR is based on; I created a new PR targeted at master; see #294

oleastre:
Is there any new release of the 0.10.5 branch foreseen ? 
If not, as #294 has been merged, this issue can be closed.

cchantep:
Next release will be of 0.11 I think

duncancrawford:
Any idea when 0.11 will be released?

sgodbillon:
@duncancrawford No ETA yet, but we're progressing quite fast, and I expect the first RC to be released less than a month from now.

sgodbillon:
Thanks @oleastre !

PR #274, opened 2015-02-24
Switched logging facilities to slf4j and pushed log4j to 'test' scope

alkersan:


cchantep:
I would prefer this to be discussed before on the [MailingList](https://groups.google.com/forum/?fromgroups#!forum/reactivemongo).

cchantep:
Thanks for your contribution. It's now [covered](https://github.com/ReactiveMongo/ReactiveMongo/blob/master/driver/src/main/scala/util/LazyLogger.scala#L19).

PR #280, opened 2015-03-03
Retry the cases of Unauthorized errors.

vvlevykin:
Use the failed authentication for retry too.

The case can be reproduced by:

1. Change the user's password.
2. Try to connect (or reconnect).
3. Change the user's password back.


vvlevykin:
Will recreate this PR: should be ReactiveMongo:0.10.5.x.akka23 instead of ReactiveMongo:master

PR #281, opened 2015-03-04
Retry authentication until it succeeds

vvlevykin:
Use the failed authentication for retry too.

For #282.

### DIFF
--- a/driver/src/main/scala/core/actors.scala
+++ b/driver/src/main/scala/core/actors.scala
@@ -421,15 +421,19 @@ class MongoDBSystem(
               None
           }
           val ns = nodeSet.updateByChannelId(response.info.channelId) { connection =>
+            val auths = nodeSet.authenticates + originalAuthenticate
             authenticateConnection(connection.copy(
               authenticated = authenticated.map(connection.authenticated + _).getOrElse(connection.authenticated),
-              authenticating = None), nodeSet.authenticates.toSeq)
+              authenticating = None), auths.toSeq)
           } { node =>
             node.copy(authenticated = authenticated.map(node.authenticated + _).getOrElse(node.authenticated))
           }
-          if (!authenticated.isDefined)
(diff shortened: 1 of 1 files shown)


vvlevykin:
Now the authentication will be retried until it eventually succeeds.
But, as there is no retry interval, it retries immediately every time and CPU usage is 100% until it succeeds.

@sgodbillon, as I understand you are the right person to ask for advice how to solve it correctly.

sgodbillon:
Hi @vvlevykin,

I think this retry should be left to the application side. As it is now, it's really hard to put an auth retry interval. But on the application side, this is quite easy to deal with: you may just `flatMap` on the future returned by `db.authenticate(user, passwd)` to create a future that "sleeps" for a little while before retrying. You can see the implementation of `Failover` to get a hint on how to that.

However, maybe we can provide a new method for this in the `DB` trait (or a new flag in the current `db.authenticate(user, passwd)` that does this automatically.)

vvlevykin:
Is there a way to do authentication on the application side when the connection is lost and then restored?

I'm asking because there was a case when MongoDB nodes were replaced, but because of misconfiguration, the credentials for one of the databases were added a bit later and the app had to be restarted to continue working.
It's certainly possible to prevent it on another level, but I thought it would be not that bad to have resilience against that.

sgodbillon:
Yes, you can call `db.authenticate(user, passwd)` again. A good way would be for you to provide a trigger (like a private rest API for example) in order to reauthenticate on demand by calling this method.

When you call `db.authenticate(user, passwd)`, if there is a successful authentication, it is stored in the connection pool metadata in order to automatically authenticate new connections. The problem is, when the authentication has failed, then it's not registered in the connection metadata. "Fixing" this should not be done, because it would mean that if did, all the new connections would automatically try to authenticate, which is likely to fail.

Issue #247, opened 2015-01-14
GroupFunctions are too restrictive


thibaultdelor:
Let's take an aggregation with a group like that : 
```
{ $group: {
            _id: null,
            favourite: { 
                $sum: { 
                    $cond:[ 
                            {$eq : ["$event", "MESSSAGE_FAVOURITED"]},
                            1,
                            -1
                        ] 
                }
            }
        }
    }
```

Because in the API Sum allow only a field or a Integer and that you can't extend GroupFunction, it is impossible to do this kind of query using the Aggregate command.
So far the only way I have found is to do it with a RawCommand (which is a pain in the ass).

I think the trait GroupFunction shouldn't be sealed, it makes everything less flexible and there is no real benefit IMO. Same apply for the other traits in aggregation.scala.

sgodbillon:
Fixed in #247

PR #287, opened 2015-03-06
Extra GroupFunction factory

cchantep:
About #247


sgodbillon:
Merging ; will mark this API deprecated in favor of the upcoming command framework.

Issue #284, opened 2015-03-05
List collection names with WiredTiger


cchantep:
When using MongoDB 3 with WiredTiger, the command `{ 'listCollections': 1 }` must be instead of `system.namespaces.find({})` to get the names of the DB collections.

Previously (with `system.namespaces`), the BSON result was as following.

```bson
{ "name" : "specs2-test-reactivemongo.geo" }
{ "name" : "specs2-test-reactivemongo.geo.$_id_" }
{ "name" : "specs2-test-reactivemongo.geo2d" }
```

Now (with `listCollections` command), the BSON result is:

```bson
{
	"cursor" : {
		"id" : NumberLong(0),
		"ns" : "specs2-test-reactivemongo.$cmd.listCollections",
		"firstBatch" : [
			{
				"name" : "geo",
				"options" : {
					
				}
			},
			{
				"name" : "geo2d",
				"options" : {
					
				}
			},
			{
				"name" : "hashed",
				"options" : {
					
				}
			},
			{
				"name" : "system.indexes",
				"options" : {
					
				}
			}
		]
	},
	"ok" : 1
}
```

PR #288, opened 2015-03-06
List collections with WiredTiger

cchantep:
Implementation for #284

### DIFF
--- a/driver/src/main/scala/api/database.scala
+++ b/driver/src/main/scala/api/database.scala
@@ -115,8 +115,12 @@ trait GenericDB[P <: SerializationPack with Singleton] { self: DB =>
 trait DBMetaCommands {
   self: DB =>
 
-  import reactivemongo.api.commands.{ Command, DropDatabase }
+  import reactivemongo.core.protocol.MongoWireVersion
+  import reactivemongo.api.commands.{
+    Command, DropDatabase, ListCollectionNames
+  }
   import reactivemongo.api.commands.bson.{ CommonImplicits, BSONDropDatabaseImplicits }
+  import reactivemongo.api.commands.bson.BSONListCollectionNamesImplicits._
   import CommonImplicits._
   import BSONDropDatabaseImplicits._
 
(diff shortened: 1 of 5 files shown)


driver/src/main/scala/api/database.scala
```diff
@@ -127,27 +127,8 @@ trait DBMetaCommands {
   /** Returns an index manager for this database. */
   def indexesManager(implicit ec: ExecutionContext) = new IndexesManager(self)
 
-  private lazy val collectionNameReader =
```
cchantep: @sgodbillon Let me know what you think about moving code in this way, to `DefaultDB` (to be able to use `runCommand` with `{ 'listCollections': 1 }`).

driver/src/main/scala/api/database.scala
```diff
@@ -140,13 +144,14 @@ trait DBMetaCommands {
...
   def collectionNames(implicit ec: ExecutionContext): Future[List[String]] = {
-    // TODO: Use command { 'listCollections': 1 }, check with WT
-    collection("system.namespaces").as[BSONCollection]()
+    val wireVer = connection.metadata.map(_.maxWireVersion)
+
+    if (wireVer.exists(_ == MongoWireVersion.V30)) {
+      Command.run(BSONSerializationPack)(self, ListCollectionNames).map(_.names)
+    } else collection("system.namespaces").as[BSONCollection]()
       .find(BSONDocument(
         "name" -> BSONRegex("^[^\\$]+$", "") // strip off any indexes
```
fehmicansaglam: Are you sure that this regex is correct?

cchantep: The tests about that on MongoDB < 3 are (still) passing with it. Have you seen any issue about?

fehmicansaglam: No, but if I am not mistaken that regex means `any name that does not contain $`. I think it should be `any name which does not start with $`.

PR #294, opened 2015-03-11
  Avoid None.get when deleting an index.

oleastre:
Same patch as PR #271 but applied to master instead of the 0.10.5.xxx branch.


sgodbillon:
Thanks!

PR #304, opened 2015-03-16
Typeclass to produce custom cursor

cchantep:

PR #307, opened 2015-03-19
Flatten strategy to go along with CursorProducer

cchantep:
Following #304 , provide flatten strategy for the type of cursor created according instance of `CursorProducer` typeclass.

### DIFF
--- a/driver/src/main/scala/api/cursor.scala
+++ b/driver/src/main/scala/api/cursor.scala
@@ -224,7 +224,7 @@ object Cursor {
   /**
    * Flattens the given future [[reactivemongo.api.Cursor]] to a [[reactivemongo.api.FlattenedCursor]].
    */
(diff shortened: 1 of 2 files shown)


driver/src/main/scala/api/cursor.scala
```diff
@@ -171,7 +171,7 @@ object Cursor {
   /**
    * Flattens the given future [[reactivemongo.api.Cursor]] to a [[reactivemongo.api.FlattenedCursor]].
    */
-  def flatten[T](future: Future[Cursor[T]]) = new FlattenedCursor(future)
+  def flatten[T, C[_] <: Cursor[_]](future: Future[C[T]])(implicit fs: CursorFlattener[C]): C[T] = fs.flatten(future)
```
cchantep: Keep the custom type of cursor across the flattening op

PR #296, opened 2015-03-11
FoldWhile operations

cchantep:
(After #301)

PR #315, opened 2015-04-07
Graceful stop for tailable cursor

cchantep:
After #296


driver/src/main/scala/api/cursor.scala
```diff
@@ -202,55 +260,94 @@ object DefaultCursor {
...
+
+    @inline
+    private def isTailable =
+      (query.flags & QueryFlags.TailableCursor) == QueryFlags.TailableCursor
+
+    /** Returns next response using tailable mode */
+    private def tailResponse(current: Response, maxDocs: Int)(implicit context: ExecutionContext): Future[Option[Response]] = mongoConnection.killed flatMap {
+      case true =>
+        logger.warn("[tailResponse] Connection is killed")
+        Future.successful(Option.empty[Response])
```
cchantep: For now, doesn't fail if connection is closed there, *just* raise that there is no more response.

PR #321, opened 2015-05-05
more efficient extraction. closes #320

### LINKED ISSUE #320: BSONDocument.getTry(key:String) is VERY expensive if key not found
https://github.com/ReactiveMongo/ReactiveMongo/blob/master/bson/src/main/scala/types.scala#L60

```
 def getTry(key: String): Try[BSONValue] = Try {
    stream.find {
      case Success(element) => element._1 == key
      case Failure(e)       => throw e
    }.map(_.get._2).getOrElse(throw DocumentKeyNotFound(key))
  }
```

Note that getOrElse throws an exception with throw DocumentKeyNotFound(key)

This may be by design, but the problem with that is that it's very costly in JVM to create an instance og exception and fill its backtrace. I was doing CPU profiling and found that this particular line costs me 5% CPU just because I don't have an attribute in the document (which is optional in my case).

I love your API but I do not understand this particular design decision: why return Try instead of Option? it's in many other places in the driver. The key benefit of this parser is macro-based .as* converters and they all go through expensive asTry path (event though asOpt is also there )

viktortnk:


cchantep:
Thanks @viktortnk

PR #326, opened 2015-05-14
Renamed variable for clarity

shirishp:
`filter` seems like a `where` clause (selection criteria), while the second parameter to find is a `select` clause (projection criteria).


shirishp:
Thanks. :)

With regards,
Shirish (http://twitter.com/_Garbage_)

PR #329, opened 2015-06-02
Adds ability to save simple name io. fully qualified

fcanedo:
`UnionType` and `AllImplementations` (in cooperation with `SaveClassName`) use the fully-qualified class name of a class to determine the run-time type of a record. The fact that the fully-qualified name is used, makes it more difficult to refactor a project’s package structure, after its initial release to production.

In order to alleviate this problem, for projects that are in full control of the package structure of the entities they are storing and want to keep their options open, this change adds “simple” versions of `SaveClassName`, `UnionTypes`and `AllImplementations` (`SaveSimpleName`, `SimpleUnionTypes` and `SimpleAllImplementations`, respectively), that do the same except that they store and match on the classes’ simple names.

### DIFF
--- a/macros/src/main/scala/interface.scala
+++ b/macros/src/main/scala/interface.scala
@@ -113,6 +113,12 @@ object Macros {
     trait SaveClassName extends Default
 
     /**
+     * Same as [[SaveClassName]] but using the class’ simple name, io.
+     * the fully-qualified name.
+     */
+    trait SaveSimpleName extends SaveClassName with Default
+
+    /**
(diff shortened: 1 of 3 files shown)


fcanedo:
Considering that the last three (open) pull requests have failing tests (that I can’t reproduce locally) in the same area, I assume that these changes are not the cause of the failing tests.

How should I proceed?

cchantep:
lgtm

cchantep:
Rebased in #360

PR #343, opened 2015-06-24
[Experimental] Removed dependency on netty plus additinal changes

sh1ng:
Hi All!

Experimental version with the following major changes
* Removed netty dependency and replaced with akka.io.tcp
* Routers to load balance requests. Plus current version uses a single actor(MongoDBSystem) to  handle request. In theory it could be a bottleneck
* Introduced actor's tree to support NodeSet, Node and Connection topology

Downside 
* Doesn't support TLS/SSL because it's not yet implemented in akka.io. It will be part of akka.stream(RC2 now).
* Akka.io can be a bottleneck for some scenarios https://groups.google.com/forum/#!topic/akka-dev/mcs5eLKiAVQ but the increase in the number of connections (MongoConnectionOptions.nbChannelsPerNode) reduces the issue.
* Removed support of mongodb 2.4
* Specs doesn't cover all cases(as well as for the current version), but all existing specs are passed.
* Some additional work require

What do you think about that changes?

Feel free to modify that branch.


cchantep:
Hi, the idea of using Akka.io can be discussed, as soon as all existing features can be implemented with.

To check it, a branch focused on this change (e.g. not with change aside, like MongoDB 2.4 deprecation) would be better.

Such branch should be (re)based on the latest master branch, so it can be reviewed.

Best regards

cchantep:
Hi, I close the ticket for now until it can have same features as netty. Then we can consider it again/reopen. Thanks your contribution, best regards.

PR #220, opened 2015-07-17
Connections per node could be set only via MongoConnectionOptions argument

cchantep:
ReactiveMongo 2.10 does not use the *nbChannelsPerNode* argument value in the **connection** method. Only *options* could be used to controll the number of connections per node.

```scala
  /**
   * Creates a new MongoConnection.
   *
   * See [[http://docs.mongodb.org/manual/reference/connection-string/ the MongoDB URI documentation]] for more information.
   *
   * @param nodes A list of node names, like ''node1.foo.com:27017''. Port is optional, it is 27017 by default.
   * @param authentications A list of Authenticates.
   * @param nbChannelsPerNode Number of channels to open per node. Defaults to 10.
   * @param name The name of the newly created [[reactivemongo.core.actors.MongoDBSystem]] actor, if needed.
   * @param options Options for the new connection pool.
   */
  def connection(nodes: Seq[String], options: MongoConnectionOptions = MongoConnectionOptions(), authentications: Seq[Authenticate] = Seq.empty, nbChannelsPerNode: Int = 10, name: Option[String] = None): MongoConnection = {
    val props = Props(new MongoDBSystem(nodes, authentications, options)())
    val mongosystem = if (name.isDefined) system.actorOf(props, name = name.get) else system.actorOf(props)
    val monitor = system.actorOf(Props(new MonitorActor(mongosystem)))
    val connection = new MongoConnection(system, mongosystem, monitor, options)
    this.synchronized {
      _connections = connection :: _connections
    }
    connection
  }
```

So, all the overloads of the **connection** method with contai the *nbChannelsPerNode* argument doesn't work properly.

### DIFF
--- a/driver/src/main/scala/api/api.scala
+++ b/driver/src/main/scala/api/api.scala
@@ -426,7 +426,7 @@ object MongoConnection {
       }
     }
 
-    def parseOptions(uriAndOptions: String): Map[String, String] = {
+    def parseOptions(uriAndOptions: String): Map[String, String] =
       uriAndOptions.split('?').toList match {
         case uri :: options :: Nil => options.split("&").map { option =>
           option.split("=").toList match {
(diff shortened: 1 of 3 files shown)


Shinbuev:
Thanks.

PR #409, opened 2015-09-25
Fix handling of @Ignore field annotation

### LINKED ISSUE #325: Ignore annotation functionality appears to be broken
@Ignore annotation is intended to skip case class parameters when generating reader and writer code using macros. I tried it an discovered that the logic is broken. What happens  is that the correspondence of parameter names and their types gets messed up. This is   evident from the following code in macros.scala:

```
    private def writeBodyConstructClass(A: c.Type): c.Tree = {
      val (constructor, deconstructor) = matchingApplyUnapply(A)
      val types = unapplyReturnTypes(deconstructor)
      val constructorParams = constructor.paramss.head

      val tuple = Ident(newTermName("tuple"))
      val (optional, required) = constructorParams.filterNot(ignoreField).zipWithIndex zip types partition (t => isOptionalType(t._2))
      val values = required map {
```

Note that filteredNot parameters are zipped with their types extracted before filtering!
Btw, there is no single unit test demonstrating @Ignore either.

avdv:
When annotating a field with `@Ignore` that field was skipped by the
macro expansion, but the index of the field was wrong, so the wrong
value was written to the BSON document.

Fixes #325

### DIFF
--- a/macros/src/main/scala/macros.scala
+++ b/macros/src/main/scala/macros.scala
@@ -185,7 +185,7 @@ private object MacroImpl {
       val constructorParams = constructor.paramss.head
 
       val tuple = Ident(newTermName("tuple"))
-      val (optional, required) = constructorParams.zipWithIndex.filterNot(p => ignoreField(p._1)) zip types partition (t => isOptionalType(t._2))
+      val (optional, required) = constructorParams.filterNot(ignoreField).zipWithIndex zip types partition (t => isOptionalType(t._2)) //constructorParams.zipWithIndex.filterNot(p => ignoreField(p._1)) zip types partition (t => isOptionalType(t._2))
       val values = required map {
         case ((param, i), typ) => {
           val neededType = appliedType(writerType, List(typ))
--- a/macros/src/main/scala/macros.scala
+++ b/macros/src/main/scala/macros.scala
@@ -185,7 +185,7 @@ private object MacroImpl {
       val constructorParams = constructor.paramss.head
 
       val tuple = Ident(newTermName("tuple"))
-      val (optional, required) = constructorParams.filterNot(ignoreField).zipWithIndex zip types partition (t => isOptionalType(t._2))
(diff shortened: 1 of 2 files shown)


cchantep:
Thanks for this contribution. It needs to be reviewed against ticket #325

avdv:
> Thanks for this contribution. It needs to be reviewed against ticket #325

Ok. But that only fixes one side of the problem: the other point is that reading a BSON document with an _ignored_ field is not defined. How to initialize those fields - `null` ?

macros/src/main/scala/macros.scala
```diff
@@ -185,7 +185,7 @@ private object MacroImpl {
       val constructorParams = constructor.paramss.head
 
       val tuple = Ident(newTermName("tuple"))
-      val (optional, required) = constructorParams.filterNot(ignoreField).zipWithIndex zip types partition (t => isOptionalType(t._2))
```
cchantep: What's the purpose of this change? The test doesn't fail with the existing code.

avdv: You must have done something wrong. Without this change the test fails with:

```
[info] x skip ignored fields
[error]    {
[error]      right: "left"
[error]    } 'BSONDocument(<non-empty>)'
[error]    
[error]     is not equal to 
[error]    
[error]    'BSONDocument(<non-empty>)' (macrospec.scala:330)
[error] Expected: ...pty>)
[error] Actual:   ...pty>)
[info] 
```

cchantep: Or *maybe* there is something wrong elsewhere with this PR. I confirm that running your tests without the change doesn't raise failure. Until the test is repeatable, this won't be merged.

avdv: > Or maybe there is something wrong elsewhere with this PR. I confirm that running your tests without the change doesn't raise failure. Until the test is repeatable, this won't be merged.

I have added a commit which reverts the change to `macros.scala`. According to Travis, the build failed for each configuration.

I don't have an explanation why the test succeeds on your machine, but maybe you should check your hardware?

avdv:
Why did you close this? Is it fixed?

PR #411, opened 2015-10-05
Implementation of 'Distinct' command

viktortnk:
http://docs.mongodb.org/manual/reference/method/db.collection.distinct/

### DIFF
--- a/driver/src/main/scala/api/collections/genericcollection.scala
+++ b/driver/src/main/scala/api/collections/genericcollection.scala
@@ -51,13 +51,17 @@ trait GenericCollectionWithCommands[P <: SerializationPack with Singleton] { sel
 }
 
 trait BatchCommands[P <: SerializationPack] {
-  import reactivemongo.api.commands.{ AggregationFramework => AC, CountCommand => CC, InsertCommand => IC, UpdateCommand => UC, DeleteCommand => DC, DefaultWriteResult, LastError, ResolvedCollectionCommand, FindAndModifyCommand => FMC }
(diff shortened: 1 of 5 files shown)


driver/src/main/scala/api/collections/genericcollection.scala
```diff
@@ -175,6 +179,16 @@ trait GenericCollection[P <: SerializationPack with Singleton] extends Collectio
    */
   def count[H](selector: Option[pack.Document] = None, limit: Int = 0, skip: Int = 0, hint: Option[H] = None)(implicit h: H => CountCommand.Hint, ec: ExecutionContext): Future[Int] = runValueCommand(CountCommand.Count(query = selector, limit, skip, hint.map(h)))
 
+  /**
```
cchantep:
```scala
/**
 * Returns the distinct values for a specified field across a single collection and returns the results in an array.
 * @param key the field for which to return distinct values
 * @param selector the query selector that specifies the documents from which to retrieve the distinct values.
 */
def distinct[T](key: String, selector: Option[pack.Document] = None)(implicit ec: ExecutionContext, reader: pack.Reader[T]): Future[immutable.SortedSet[T]] = ???
```

cchantep:
Hi, thanks for your contribution. It will be reviewed soon.

cchantep:
Please squash the commit into a single one for the final review. Thanks

viktortnk:
@cchantep Done

cchantep:
Thanks @viktortnk

PR #423, opened 2015-11-17
Stop immediately without active connections

avdv:
When a Close message is received, the MongoDBSystem switches into the
`closing` state and gently closes down all remaining connections.

All monitors are notified once all connections are closed. This depends
on at least a single ChannelClosed message to be received by the
`closing` receive method. Alas, this message will never arrive when
there're no active connections.

This will ultimately result in a timeout exception after a 10 seconds
timeout has elapsed.

For us, this is especially a problem when we want to restart our Play application which depends on the reactivemongo module. It takes at least 10 seconds before the application shuts down in case there is no active connection. Test runs suffer to, because even when we don't need use the Controller depending on the MongoDB connection, it still takes 10 seconds to shut down the FakeApplication.

### DIFF
--- a/driver/src/main/scala/core/actors.scala
+++ b/driver/src/main/scala/core/actors.scala
@@ -193,6 +193,25 @@ trait MongoDBSystem extends Actor {
     connections = connections,
     authenticated = if (connections.isEmpty) Set.empty else node.authenticated)
 
+  def stopWhenDisconnected(state: String, msg: AnyRef): Unit = {
+    val remainingConnections = nodeSet.nodes.foldLeft(0)(
+      { (open, node) => open + node.connections.size })
+
+    if (logger.logger.isDebugEnabled()) {
+      val disconnected = nodeSet.nodes.foldLeft(0) { (open, node) =>
+        open + node.connections.count(_.status == ConnectionStatus.Disconnected)
+      }
+
+      logger.debug(s"(State: $state) Received $msg remainingConnections = $remainingConnections, disconnected = $disconnected, connected = ${remainingConnections - disconnected}")
+    }
+
+    if (remainingConnections == 0) {
+      monitors foreach (_ ! Closed)
+      logger.info(s"MongoDBSystem $self is stopping.")
+      context.stop(self)
+    }
+  }
+
   def updateNodeSetOnDisconnect(channelId: Int): NodeSet =
     updateNodeSet(nodeSet.updateNodeByChannelId(channelId) { node =>
       val connections = node.connections.map { connection =>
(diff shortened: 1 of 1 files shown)


cchantep:
Hi, letting aside tests are failing there, I don't think this is the way to go. For if the issue is when testing with Play `FakeApplication`, which load/unload the plugin/pool each time, this fix should be at this level.

avdv:
> Hi, letting aside tests are failing there

Seems to be a build server problem, right?

> I don't think this is the way to go.

Care to explain? I already asked you about this, you said go for it (see https://groups.google.com/forum/?fromgroups#!topic/reactivemongo/sMaZUoiOA30) and I think the fix is reasonable. What did you have in mind instead?

> For if the issue is when testing with Play FakeApplication, which load/unload the plugin/pool each time, this fix should be at this level.

No, the real issue is in _production_, when we need to restart the application, it takes 10 seconds until the application closes, just waiting for a message that is never going to happen.

cchantep:
>Seems to be a build server problem, right?

I don't think so.

>> I don't think this is the way to go.
>
>Care to explain? I already asked you about this, you said go for it
>(see
>https://groups.google.com/forum/?fromgroups#!topic/reactivemongo/sMaZUoiOA30)
>and I think the fix is reasonable. What did you have in mind instead?
>
>> For if the issue is when testing with Play FakeApplication, which
>load/unload the plugin/pool each time, this fix should be at this
>level.
>
>No, the real issue is in _production_, when we need to restart the
>application, it takes 10 seconds until the application closes, just
>waiting for a message that is never going to happen.

I don't see the case where a Play app would be programmatically restarted (keeping the same JVM).

The current driver or plugin should close, whenever the it can be properly, after the close timeout.

avdv:
>> Seems to be a build server problem, right?
> I don't think so.

There _definitively_ are problems on/with the CI system. I triggered a re-run and all jobs passed with flying colors, except for the three that failed (again) because of connection problems to the Mongo DB _before even starting the build_.

> > I don't think this is the way to go.

Why? It's a simple change that fixes the current behavior.

Note: there's actually no change in behavior whatsoever for the `closing` state. In the `processing` state, there's only a single additional check for the number of remaining connections when receiving the `Stop` message. And only _if there are no connections_ left, the same thing is done as in the `closing` state: notify all monitors and stop the context.

Alternatively, in you opinion, what would be the way to go?

>> No, the real issue is in _production_, when we need to restart the
>> application, it takes 10 seconds until the application closes, just
>> waiting for a message that is never going to happen.
> I don't see the case where a Play app would be programmatically restarted (keeping the same JVM).

You misunderstood, we want to restart the JVM, like `systemctl restart playapp`. But it doesn't really matter how you stop it, there's always the 10 second delay and then the TImeoutException.

> The current driver or plugin should close, whenever the it can be properly, after the close timeout.

Yes, it should. But it doesn't do it "properly", it exits with a timeout exception _when_ there are no active connections to the MongoDB (for whatever reason).

Do you at least admit the problem here? 10 seconds additional downtime because of no reason is unacceptable, no?

cchantep:
You can have a look at https://github.com/cchantep/ReactiveMongo/commit/00b7a6ef333f822c64eb429a7eda1fa251785c2b

PR #424, opened 2015-11-28
add mongodb 3.2 $sample aggregation pipeline operator

ornicar:
already in use on lichess.org


cchantep:
Thanks @ornicar . Could you amend with a test in [AggregationSpec](https://github.com/ReactiveMongo/ReactiveMongo/blob/master/driver/src/test/scala/AggregationSpec.scala) before we merge?

cchantep:
@ornicar merged as #428

PR #459, opened 2016-02-19
Support for SocketTimeoutMS on URL connection

charleskubicek:
Implements semantics described in https://docs.mongodb.org/manual/reference/connection-string/#uri.socketTimeoutMS

### DIFF
--- a/driver/src/main/scala/api/MongoConnectionOptions.scala
+++ b/driver/src/main/scala/api/MongoConnectionOptions.scala
@@ -24,6 +24,7 @@ case object ScramSha1Authentication extends AuthenticationMode
  * @param nbChannelsPerNode Number of channels (connections) per node (ReactiveMongo-specific option).
  * @param writeConcern the default write concern
  * @param readPreference the default read preference
+ * @param socketTimeoutMS The number of milliseconds a socket can be idle before it is closed
  */
 case class MongoConnectionOptions(
   // canonical options - connection
@@ -41,4 +42,5 @@ case class MongoConnectionOptions(
 
   // read and write preferences
   writeConcern: WriteConcern = WriteConcern.Default,
-  readPreference: ReadPreference = ReadPreference.primary)
+  readPreference: ReadPreference = ReadPreference.primary,
+  socketTimeoutMS: Int = 0)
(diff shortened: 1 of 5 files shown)


cchantep:
Hi, please squash the commits to a single one, so it can be reviewed. Best regards.

charleskubicek:
Squashed commits into one.

cchantep:
Thanks. It seems to break to timeout handling (see on Travis).

cchantep:
Please rebase against master, to check whether there is any conflict.

cchantep:
Please make sure to have run scalariform (`sbt scalariformFormat test:scalariformFormat`), so the coding style is not broken.

cchantep:
Hi, the tests are needed. I prefer to wait they are fixed, rather than removing them. Thanks.

charleskubicek:
Hi @cchantep, I have rebased. I have also temporarily removed the one test I added called "Connection should timeout when it is idle". This appeared to cause some other tests to fail for an unknown reason. The tests are now passing on Travis (except the JVM crashes). 

I wasn't happy with that test as it was too heavy-weight, can you think of a better way to test SocketTimeoutMS is correctly applied? Or, do we need to? All it does is delegate to Netty, but then we should unit test MongoHandler.

cchantep:
@charleskubicek Why this test (which should be isolated) was slowing the other tests?

cchantep:
Hi @charleskubicek, can it be tested using `netem` at first?

charleskubicek:
Hi @cchantep, thanks for the tip, I'll take a look. We have been testing using a proxy service between the driver and Mongo to simulate slowdowns to test the idle trigger - so far it works well.

But I can't use `netem` in automated tests as it's not cross platform. We're going to look at other ways of unit testing it this week.

cchantep:
As long as you can reproduce the test env using `netem`, that's ok, as it's already used for other integration tests. Let me know when you have something working in this way (and please rebase to keep up-to-date). Best regards.

cchantep:
Hi, reintroduced an embedded [test proxy](https://github.com/ReactiveMongo/ReactiveMongo/blob/master/driver/src/test/scala/common.scala#L92). You may want to have a look at.

PR #497, opened 2016-04-03
Support for socketTimeoutMS

cchantep:
Fix #459


cchantep:
@charleskubicek Back to that, I'm now thinking that rather corresponds to [maxIdleTimeMS](https://docs.mongodb.com/manual/reference/connection-string/#urioption.maxIdleTimeMS) in the MongoDB terminology, the `socketTimeoutMS` being "covered" by the failover in an asynchronous way.

PR #509, opened 2016-06-04
Support taggable read preference on shard router

cchantep:
ReactiveMongo Version: 0.11.7
MongoDB version: 2.6.9, 3.x

Expected Behavior
1. Configure mongo driver to connect to mongos node
2. Configure ReadPreference.secondaryPreferred(BSONDocument("dc"->"sw"))
3. Execute find operation with the explain flag
4. The nodes with tag {"dc"->"sw"} should reply

Actual Behavior
4. mongos selects the nodes for processing request without the tag information. Just slaveOk flag is used.

From the source code of ReactiveMongo driver tags are used to create a partial function and perform the matching operation on client side. Matching logic is not performed when connection is established to mongos node (source code: nodeset.scala):
```
  private def pickFromGroupWithFilter(roundRobiner: RoundRobiner[Node, Vector], filter: Option[BSONDocument => Boolean], fallback: => Option[Node]) =
    filter.fold(fallback)(f =>
      roundRobiner.pickWithFilter(_.tags.fold(false)(f)))

  // http://docs.mongodb.org/manual/reference/read-preference/
  def pick(preference: ReadPreference): Option[(Node, Connection)] = {
    if (mongos.isDefined) {
      pickConnectionAndFlatten(mongos)
    } else preference match {
      case ReadPreference.Primary                    => pickConnectionAndFlatten(primary)
      case ReadPreference.PrimaryPreferred(filter)   => pickConnectionAndFlatten(primary.orElse(pickFromGroupWithFilter(secondaries, filter, secondaries.pick)))
      case ReadPreference.Secondary(filter)          => pickConnectionAndFlatten(pickFromGroupWithFilter(secondaries, filter, secondaries.pick))
      case ReadPreference.SecondaryPreferred(filter) => pickConnectionAndFlatten(pickFromGroupWithFilter(secondaries, filter, secondaries.pick).orElse(primary))
      case ReadPreference.Nearest(filter)            => pickConnectionAndFlatten(pickFromGroupWithFilter(nearestGroup, filter, nearest))
    }
  }
```

In mongo-java driver read preference is a part of the wire protocol, so mongos is able to select the nodes based on the tags (source code: FindOperation.java):
```
private BsonDocument asDocument(final ConnectionDescription connectionDescription, final ReadPreference readPreference) {
    BsonDocument document = new BsonDocument();

    if (modifiers != null) {
        document.putAll(modifiers);
    }

    if (sort != null) {
        document.put("$orderby", sort);
    }

    if (maxTimeMS > 0) {
        document.put("$maxTimeMS", new BsonInt64(maxTimeMS));
    }

    if (connectionDescription.getServerType() == SHARD_ROUTER && !readPreference.equals(primary())) {
        document.put("$readPreference", readPreference.toDocument());
    }

    if (document.isEmpty()) {
        document = filter != null ? filter : new BsonDocument();
    } else if (filter != null) {
        document.put("$query", filter);
    } else if (!document.containsKey("$query")) {
        document.put("$query", new BsonDocument());
    }

    return document;
}
```

The tags from read preference are not propagated to the request
bsoncollection.scala
```
  def merge(readPreference: ReadPreference): BSONDocument = {
    // Primary and SecondaryPreferred are encoded as the slaveOk flag;
    // the others are encoded as $readPreference field.
    val readPreferenceDocument = readPreference match {
      case ReadPreference.Primary                    => None
      case ReadPreference.PrimaryPreferred(filter)   => Some(BSONDocument("mode" -> "primaryPreferred"))
      case ReadPreference.Secondary(filter)          => Some(BSONDocument("mode" -> "secondary"))
      case ReadPreference.SecondaryPreferred(filter) => None
      case ReadPreference.Nearest(filter)            => Some(BSONDocument("mode" -> "nearest"))
    }
    val optionalFields = List(
      sortOption.map { "$orderby" -> _ },
      hintOption.map { "$hint" -> _ },
      maxTimeMsOption.map { "$maxTimeMS" -> BSONLong(_) },
      commentString.map { "$comment" -> BSONString(_) },
      option(explainFlag, "$explain" -> BSONBoolean(true)),
      option(snapshotFlag, "$snapshot" -> BSONBoolean(true)),
      readPreferenceDocument.map { "$readPreference" -> _ }).flatten
    val query = queryOption.getOrElse(BSONDocument())
    if (optionalFields.isEmpty)
      query
    else
      BSONDocument(("$query" -> query) :: optionalFields)
  }
```

### DIFF
--- a/driver/src/main/scala/api/ReadPreference.scala
+++ b/driver/src/main/scala/api/ReadPreference.scala
@@ -7,44 +7,185 @@ import reactivemongo.bson.{ BSONArray, BSONDocument, BSONDocumentWriter }
  * with a predefined strategy.
  */
 sealed trait ReadPreference {
+  /** Indicates whether a slave member is ok. */
   def slaveOk: Boolean = true
+
   def filterTag: Option[BSONDocument => Boolean]
 }
 
 object ReadPreference {
-  /** Read only from the primary. This is the default choice. */
+  /** Reads only from the primary. This is the default choice. */
   object Primary extends ReadPreference {
-    override def slaveOk = false
-    override def filterTag = None
+    override val slaveOk = false
+    val filterTag = None
 
-    override def toString = "Primary"
+    override val toString = "Primary"
   }
 
-  /** Read from the primary if it is available, or secondaries if it is not. */
-  case class PrimaryPreferred(
-      filterTag: Option[BSONDocument => Boolean]) extends ReadPreference {
-    override def toString = s"PrimaryPreferred($filterTag)"
+  private def TagFilter(tagSet: Seq[Map[String, String]]): Option[BSONDocument => Boolean] = if (tagSet.isEmpty) None else Some { doc: BSONDocument =>
+    val matching = tagSet.find(_.foldLeft(Map.empty[String, String]) {
+      case (ms, (k, v)) =>
+        doc.getAs[String](k).filter(_ == v) match {
+          case None => ms + (k -> v)
+          case _    => ms
+        }
+    }.isEmpty)
+
+    matching.isDefined
   }
 
-  /** Read only from any secondary. */
-  case class Secondary(
-      filterTag: Option[BSONDocument => Boolean]) extends ReadPreference {
-    override def toString = s"Secondary($filterTag)"
+  private def TagFilter(tagSet: Seq[BSONDocument]): Option[BSONDocument => Boolean] = if (tagSet.isEmpty) None else Some { doc =>
+    tagSet.exists(doc.contains(_))
   }
 
-  /** Read from any secondary, or from the primary if they are not available. */
-  case class SecondaryPreferred(
-      filterTag: Option[BSONDocument => Boolean]) extends ReadPreference {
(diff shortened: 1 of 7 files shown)


Shinbuev:
Thanks Cédric that you has paid an attention to this issue.
I'm a little bit confused and think that there are few issues in the code. Please, tell me if I'm wrong:
```
/** Reads from any node that has the given `tagSet` in the replica set (preferably a secondary). */
def secondaryPreferred[T](tagSet: T*)(implicit writer: BSONDocumentWriter[T]): PrimaryPreferred = new PrimaryPreferred(TagFilter(tagSet.map(writer.write(_))))
```
PrimaryPreferred now is using for all the types of read preference: nearest, secondary, secondaryPreferred.

_tags private fields are not initialized/copied in the constructor/apply/copy method for read preference.

There is ReadPreferenceSpec.scala in my branch that might help to cover the cases with creation of read preference

cchantep:
My bad. Has been fixed with 0.11.13.

PR #519, opened 2016-06-07
This fixes ReadPreference factory method: use private constructor with the list of maps.

Shinbuev:
Fixes:
primaryPreferred(List(Map))
secondaryPreferred(List(Map))
secondary(List(Map)
nearest(List(Map))


Shinbuev:
Would you take look at usage of private constructor from the tagSet: List(Map[String, String]) in read preference classes

cchantep:
Fixed by #521

PR #524, opened 2016-06-26
BsonDocument with binary is mutated on reads

cchantep:
Scala version 2.11
ReactiveMongo version 0.11.13 (current)
MongoDB version - not relevant
OS - Windows 7
JDK - 1.8.0_92

Reading a binary element in a BSonDocument consumes the binary data and mutates the object. Reading elements of other types do not consume them so this is unexpected and undesired behavior, nor do I see this behavior documented. The following test should pass.

```scala
  test("bug report") {
    val dbObject = BSONDocument("d" -> Array[Byte](1, 2, 3))
    dbObject.getAs[Array[Byte]]("d").get.contains(1.toByte) shouldBe true  // passes
    dbObject.getAs[Array[Byte]]("d").get.contains(1.toByte) shouldBe true  // fails because d was consumed
  }
```

fails with
```
None.get
java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:347)
	at scala.None$.get(Option.scala:345)
	at BugTest$$anonfun$6.apply$mcV$sp(BugTest.scala:8)
```

### DIFF
--- a/bson/src/main/scala/handlers.scala
+++ b/bson/src/main/scala/handlers.scala
@@ -236,7 +236,7 @@ trait DefaultBSONHandlers {
   }
 
   implicit object BSONBinaryHandler extends BSONHandler[BSONBinary, Array[Byte]] {
-    def read(bin: BSONBinary) = bin.value.readArray(bin.value.size)
+    def read(bin: BSONBinary) = bin.value.duplicate().readArray(bin.value.size)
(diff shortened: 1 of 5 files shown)


cchantep:
Fixed in [0.11.14](http://search.maven.org/#artifactdetails%7Corg.reactivemongo%7Creactivemongo_2.11%7C0.11.14%7Cjar)

JeffAtDeere:
Awesome! Thank you

PR #426, opened 2016-07-19
Lookup agg stage

cchantep:


cchantep:
You can use if from the 0.12.0-SNAPSHOT. A RC0 is coming.

PR #592, opened 2016-09-27
Override ReadPreference on collection reference

cchantep:


cchantep:
@ornicar `coll.withReadPreference(..)` to specify preference for a coll ref, and thus ops applied on it

PR #601, opened 2016-11-03
Update to Scala 2.12.0

cchantep:


jbgi:
From what I understand, the major blocker for this update is [play-iteratees](https://github.com/playframework/play-iteratees) which have not been published against 2.12. On the other hand it seams that `play-iteratees` is being deprecated in favour of `akka-stream` and has not seen any update for months.

Can ReactiveMongo drop support for `play-iteratees` altogether?

cchantep:
That's not the blocker

alari:
@cchantep when are you going to make 0.12 release for scala 2.12? Is there any kind of snapshot to work with?

cchantep:
When the required dependencies themselves will be available for 2.12.

PR #606, opened 2016-12-13
SSL/TLS Client Authentication

opetch:
This PR adds support mutual SSL/TLS authentication, by adding support for client certificates.

This will work with a JKS keystore configured with the JDK keystore property "javax.net.ssl.keyStore". It is certainly possible to make this more configurable or to change the default behaviour. I have implemented this as my current client need it and it has been implemented to their requirements. They in fact have their own fork of ReactiveMongo where I originally implemented this, however I've  also added it here hoping that the community may benefit from it.

I have tested this on my own Play applications also by running the test suite with SSL enabled and keystore setup, against a mongod instance with the --sslCAFile attribute set and the tests pass. The suite fails in this configuration without these changes.


cchantep:
Hi,

Thx for your contribution. Aside from the pending review, could you please suggest a scenario for integration test about that.

driver/src/main/scala/core/nodeset.scala
```diff
@@ -667,6 +663,41 @@ final class ChannelFactory private[reactivemongo] (
...
+      val password = Option(System.getProperty("javax.net.ssl.keyStorePassword")).getOrElse("")
+
+      val ks = {
+        val res = KeyStore.getInstance("JKS")
+        val fin = new FileInputStream(path)
+        res.load(fin, password.toCharArray)
+        res
+      }
+
+      val algorithm = Option(Security.getProperty("ssl.KeyManagerFactory.algorithm")).getOrElse("SunX509")
```
cchantep: Need to update the documentation: https://github.com/ReactiveMongo/reactivemongo-site/blob/gh-pages/releases/0.12/documentation/tutorial/connect-database.md

opetch: In fact I can change to code to call getDefaultAlgorithm() as it should do the same thing, serves me right for not looking into the API and just copying some example code, will update this in the next push.

cchantep: It must be documented that it relies on JKS, so people can at least have a look at its specific documentation

opetch:
As you can see I've done the code review comments, I must have been tired when I merged my changes with master as I failed to notice that I'd called init on the ctx twice and so did you ;P anyway hope this is better

opetch:
As for the integration test I guess we should alter the travis scripts to start mongo up without --sslAllowInvalidCertificates and with --sslCAFile and then ensure we configure the suite to send certs with the relevant keystore bits added. I'll take a look at this in the next few days

opetch:
I've added a couple of tidy up commits now, do you want me to squash them and then force push to my fork?

driver/src/main/scala/core/nodeset.scala
```diff
@@ -663,16 +667,20 @@ final class ChannelFactory private[reactivemongo] (
...
-  private def sslContext = {
+  private def sslContext(tm: Array[javax.net.ssl.TrustManager]) = {
+    import java.io.FileInputStream
+    import java.security.{ KeyStore, Security }
+    import javax.net.ssl.KeyManagerFactory
 
-    val keyManagers: Array[KeyManager] = Option(System.getProperty("javax.net.ssl.keyStore")).map { path =>
+    val keyManagers = Option(System.getProperty("javax.net.ssl.keyStore")).map { path =>
 
       val password = Option(System.getProperty("javax.net.ssl.keyStorePassword")).getOrElse("")
```
cchantep: `.orNull` ?

driver/src/main/scala/core/nodeset.scala
```diff
@@ -663,16 +667,20 @@ final class ChannelFactory private[reactivemongo] (
...
 
       val ks = {
         val res = KeyStore.getInstance("JKS")
         val fin = new FileInputStream(path)
         res.load(fin, password.toCharArray)
+        fin.close()
```
cchantep: nope, that's not a valid resource management

driver/src/main/scala/core/nodeset.scala
opetch: I'm using that orNull now a bit further down inside the call to init, must of pushed that just after you commented here

driver/src/main/scala/core/nodeset.scala
opetch: ok it is now, I was being a bit cheeky/cowboyish and assuming that you would only hit an IOException in the event of misconfiguration and then your not going to be running the application for long anyway as it won't work. But I've put the close in a finally block now, thought about doing something fancier such as a scala function implementing try with resources but decided to keep it simple.

driver/src/main/scala/core/nodeset.scala
```diff
@@ -1,19 +1,34 @@
...
+import java.util.concurrent.{ TimeUnit, Executor, Executors }
 
 import scala.collection.generic.CanBuildFrom
 import scala.collection.immutable.Set
+
 import shaded.netty.util.HashedWheelTimer
 import shaded.netty.buffer.HeapChannelBufferFactory
 import shaded.netty.channel.socket.nio.NioClientSocketChannelFactory
-import shaded.netty.channel.{ Channel, ChannelFuture, ChannelPipeline, Channels }
+import shaded.netty.channel.{
```
cchantep: Please don't reformat existing imports

driver/src/main/scala/core/nodeset.scala
```diff
@@ -667,6 +655,53 @@ final class ChannelFactory private[reactivemongo] (
...
   }
 
+  private def sslContext = {
+    import java.io.FileInputStream
+    import java.security.KeyStore
+    import javax.net.ssl.{ KeyManagerFactory, TrustManager }
+
+    val keyManagers = Option(System.getProperty("javax.net.ssl.keyStore")).map { path =>
+
+      val password = Option(System.getProperty("javax.net.ssl.keyStorePassword")).getOrElse("")
```
opetch: I think that's me correcting them back to what it should be, see overall changes between my branch and master

driver/src/main/scala/core/nodeset.scala
opetch: I think that's me correcting them back to what it should be, see overall changes between my branch and master

driver/src/main/scala/core/nodeset.scala
cchantep: the indent is wrong

driver/src/main/scala/core/nodeset.scala
cchantep: Would be better to disable that option in your IDE while editing there

opetch: yeah I know, it's useful for most projects I commit on, I'll disable it so it doesn't happen on reactive mongo

driver/src/main/scala/core/nodeset.scala
opetch: Any chance of a response? Sorry to be pushy, it's just you've been so quick to respond so far so thought you may have missed my question. Would be fantastic to get this merged, I'm going to start looking at the integration testing this evening.

cchantep: Seems ok now. BTW it won't get merged before testing.

opetch: yep that's understandable, will get back to you on that

opetch: Nice job :+1: I had a little look into the scripts but hadn't had the time to implement anything so thanks for getting this started. I've merged your commit in to my fork, today I'll have a go at the TODOs.

opetch: That's all done now, mutual-ssl and self-ssl jobs are passing, a couple have timed out but I'm restarting them

opetch: In fact I can't restart the jobs in the build referenced below, as I'm not a maintainer, but I did do so on my own forks's Travis and all is passing....
https://travis-ci.org/opetch/ReactiveMongo/builds/185446959

.ci_scripts/beforeInstall.sh
```diff
@@ -30,21 +32,25 @@ fi
 
 echo "Max connection: $MAX_CON"
 
-# OpenSSL
```
cchantep: I prefer that to be reverted so the binary is the same in any case, even if the options are differents

.ci_scripts/beforeInstall.sh
```diff
@@ -84,14 +90,14 @@ else
     echo "  maxIncomingConnections: $MAX_CON" >> /tmp/mongod.conf
 fi
 
-mkdir /tmp/mongodb
+mkdir -p /tmp/mongodb
```
cchantep: no need

.ci_scripts/beforeInstall.sh
```diff
@@ -84,14 +90,14 @@ else
...
 
 SSL_PASS=""
 
 if [ "$MONGO_PROFILE" = "self-ssl" -o "$MONGO_PROFILE" = "mutual-ssl" ]; then
-    SSL_PASS=`uuidgen`
+    SSL_PASS="secret"
```
cchantep: prefer the generated/random

.ci_scripts/beforeInstall.sh
```diff
@@ -110,7 +116,11 @@ EOF
     CAFile: $SCRIPT_DIR/client.pem
 EOF
 
-        # TODO: JKS https://blog.codecentric.de/en/2013/01/how-to-use-self-signed-pem-client-certificates-in-java/
+        keytool -importkeystore  -srcstoretype PKCS12 \
```
cchantep: indent issue

.ci_scripts/beforeInstall.sh
opetch: Sure I can revert it, it's just the compilation takes some time and it slows down the suites that don't need it.

cchantep: The compilation is cached on TravisCI

.ci_scripts/beforeInstall.sh
cchantep: What's the error ?

.ci_scripts/beforeInstall.sh
cchantep: The indent is 4-spaces. Please use that.

driver/src/main/scala/core/nodeset.scala
opetch: having to revert back to a static password as uuidgen not found
https://travis-ci.org/opetch/ReactiveMongo/jobs/185468041

.ci_scripts/beforeInstall.sh
```diff
@@ -91,7 +91,7 @@ mkdir /tmp/mongodb
 SSL_PASS=""
 
 if [ "$MONGO_PROFILE" = "self-ssl" -o "$MONGO_PROFILE" = "mutual-ssl" ]; then
-    SSL_PASS=`uuidgen`
+    SSL_PASS=secret
```
cchantep: What's the issue there?

cchantep: `git rev-parse HEAD`

opetch: If we really do want to generate something on the fly I've tested the following on the traivs docker container and it works
`date | sha1sum | cut -f 1 -d ' '`

cchantep: ok

opetch: in fact so will `cat /proc/sys/kernel/random/uuid` so I'll switch to that

.ci_scripts/beforeInstall.sh
opetch: -p stops the command failing if the dir exists, as said useful when re-running the script in the docker container, anyway it's gone now

.ci_scripts/beforeInstall.sh
opetch: ah nice, now I understand the caching config!

opetch:
@cchantep is it worth increasing the patience config for waiting on futures? Getting a lot of jobs that fail and pass after a restart because it's set to 2 seconds

cchantep:
no, it should pass with the same env

opetch:
Not sure I understand why we can't increase the 2 seconds, is it not just an arbitrary amount of time which happens to be sufficient for most but not all job runs? 

Anyway not a big issue, all my jobs https://travis-ci.org/opetch/ReactiveMongo/builds/185482940 are passing now. Is this ready for merge then?

cchantep:
The Travis CI must first validates it

opetch:
@cchantep can you restart this job please?
https://travis-ci.org/ReactiveMongo/ReactiveMongo/jobs/185483041

opetch:
ready to merge?

cchantep:
@opetch you need first to go to [contributor agreement](https://www.clahub.com/agreements/ReactiveMongo/ReactiveMongo).

opetch:
What do I need to do to get this to pass, the only text I can read for the fail is...

"One or more of this commit's parents has co..."

I assume this is because I merged a commit made on HMRC/ReactiveMongo, this fork still has the same license and copyright under "Stephane Godbillon (@sgodbillon) and Zenexity"

cchantep:
The reason is to have a explicit contributor agreement: even if the license is nothing new and involve an implicit agreement of the related terms for all the past contributions, it's better to have the contributors since now giving an explicit agreement.

To do so you can a blue button "I agree" after the form.

opetch:
Where is this blue button? I have signed the CLA, but the check still fails

opetch:
@cchantep Would be nice to restart the failed jobs (I made the keystore type configurable) and resolve the issue with the CLA, is claHub integration broken? Should it be passing if I've agreed?

opetch:
@cchantep can we restart the failing jobs again? also I sent you a msg on gitter, I assume that will get to you?

opetch:
@cchantep thanks for merging this, bye the way did you see my message above re: gitter. If you don't use it should I email you?

cchantep:
Gitter is not used for ReactiveMongo, the [MailingList](https://groups.google.com/forum/?fromgroups#!forum/reactivemongo) is available.

PR #600, opened 2017-01-05
Macros.handler vs Macros.handleOpts for sealed trait

cchantep:
``` scala
sealed trait Foo
case object Bar extends Foo
case object Lorem extends Foo

import reactivemongo.bson.{ Macros, BSONDocument }, BSONDocument.pretty

val h1 = Macros.handler[Foo]
val h2 = Macros.handlerOpts[Foo, Macros.Options.AllImplementations]

pretty(h1.write(Bar))
/*
{

}
*/

pretty(h2.write(Bar))
/*
{
  className: "Bar"
}
*/
```


cchantep:
cc @maxime-gautre

PR #617, opened 2017-01-12
implement mongo authentication for x509 certificates

### LINKED ISSUE #462: X509 authentication
https://docs.mongodb.org/manual/core/security-x.509/#security-auth-x509

opetch:
# Pull Request Checklist
 
* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)? yes
* [x] Have you read through the [contributor guidelines](https://github.com/ReactiveMongo/ReactiveMongo/blob/master/CONTRIBUTING.md#reactivemongo-developer--contributor-guidelines)? yes
* [x] Have you [squashed your commits](https://www.playframework.com/documentation/2.4.x/WorkingWithGit#Squashing-commits)? only one anyway
* [ ] Have you added tests for any changed functionality? not yet

## Fixes

Fixes #462 

Hi @cchantep I've implemented auth for X509 will be adding tests to DriverSpec but wanted to get your opinion on the implementation earlier. Also thinking another configuration in the travis matrix would be a good idea, that will give it more thorough coverage, what do you think?

I'll also be updating the docs

### DIFF
--- a/driver/src/main/scala/api/api.scala
+++ b/driver/src/main/scala/api/api.scala
@@ -44,7 +44,6 @@ import reactivemongo.core.actors.{
   Closed,
   Exceptions,
   ExpectingResponse,
-  LegacyDBSystem,
   MongoDBSystem,
   PrimaryAvailable,
   PrimaryUnavailable,
@@ -722,6 +721,9 @@ object MongoConnection {
           case ("authSource", v) => unsupported -> result.
             copy(authSource = Some(v))
 
+          case ("authMode", "x509") => unsupported -> result.
+            copy(authMode = X509Authentication)
+
           case ("authMode", "mongocr") => unsupported -> result.
             copy(authMode = CrAuthentication)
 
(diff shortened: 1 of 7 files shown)


driver/src/main/scala/api/api.scala
```diff
@@ -952,16 +954,21 @@ class MongoDriver(
    */
   def connection(nodes: Seq[String], options: MongoConnectionOptions = MongoConnectionOptions(), authentications: Seq[Authenticate] = Seq.empty, name: Option[String] = None): MongoConnection = {
     val nm = name.getOrElse(s"Connection-${+MongoDriver.nextCounter}")
+    import reactivemongo.core.actors.{ MongoScramSha1Authentication, MongoCrAuthentication, MongoX509Authentication }
 
     // TODO: Passing ref to MongoDBSystem.history to AddConnection
     lazy val dbsystem: MongoDBSystem = options.authMode match {
-      case CrAuthentication => new LegacyDBSystem(
+      case CrAuthentication => new StandardDBSystem(
```
cchantep: This must not changed : Mongo-CR is legacy

driver/src/main/scala/core/actors.scala
```diff
@@ -1245,29 +1245,13 @@ trait MongoDBSystem extends Actor {
 }
 
 @deprecated("Internal class: will be made private", "0.11.14")
-final class LegacyDBSystem private[reactivemongo] (
```
cchantep: I prefer not to mix deprecation cleanup and feature in the same PR. For now `LegacyDBSystem` must be kept as-is for compatibility.

driver/src/main/scala/core/nodeset.scala
```diff
@@ -505,16 +507,14 @@ case class CrAuthenticating(db: String, user: String, password: String, nonce: O
 }
 
 case class ScramSha1Authenticating(
-    db: String, user: String, password: String,
```
cchantep: Take care to the formatting

driver/src/main/scala/core/actors.scala
opetch: I was trying to avoid having to define 3 DbSystems but no problem if you'd prefer that. Any ideas for a name X509DbSystem?

cchantep: Or a trait mix-in with Standard

driver/src/main/scala/core/nodeset.scala
opetch: That should be fixed, I noticed just after raising the PR then amended the commit and forced pushed to overwrite it, but you were too quick :)
Anyway think Github might be confused (because of the force push) as it is still showing these formatting changes

driver/src/main/scala/core/actors.scala
opetch: Ah got you, just put LegacyDbSystem back in

githubmo:
Hi @cchantep, I'm looking to pick this up from where it was left off as we need this done. So that we are on the same page, is there any suggestions you have for functional tests that would satisfy the 4th point of the PR and get this merged? @opetch, any chance you have tests that weren't committed that you can push up?

cchantep:
https://github.com/ReactiveMongo/ReactiveMongo/pull/606

githubmo:
@cchantep #606 is for TLS/SSL connections. I am specifically talking about certificate authorisation, which is a separate concern that this PR is trying to address. https://docs.mongodb.com/manual/tutorial/configure-x509-client-authentication/#use-x-509-certificates-to-authenticate-clients

cchantep:
See #722

PR #623, opened 2017-01-19
Make WriteResult.lastError detect/return any error. Fix #622

### LINKED ISSUE #622: Prefer structured errors on ordered bulk command failures
While doing ordered bulk inserts I noticed that an error on a individual document get lost into a `GenericDriverException`.
I'd rather have a more structured `LastError`. (because I need to recover from duplicate key errors)

The reason of the behavior is that in `reactivemongo.api.collections.GenericCollection.Mongo26WriteCommand#send` we have:
```scala
       case Some(wr) if (wr.inError || (wr.hasErrors && ordered)) => {
          Future.failed(WriteResult.lastError(wr).
            getOrElse[Exception](GenericDriverException(
              s"write failure: $wr"
            )))
        }
```
and in `reactivemongo.api.commands.WriteResult#lastError`:
```scala
  def lastError(result: WriteResult): Option[LastError] = result match {
    case error @ LastError(_, _, _, _, _, _, _, _, _, _, _, _, _, _) => Some(error)
    case _ if (result.ok) => None
    case _ => Some(LastError(
      false, // ok
      result.errmsg,
```
Eg, if mongo (tested with 3.2) fail to insert a document due to a duplicate key, it will return an response like:
```
{
  ok: BSONInteger(1),
  n: BSONInteger(0),
  writeErrors: [
    0: {
      index: BSONInteger(0),
      code: BSONInteger(11000),
      errmsg: "E11000 duplicate key error collection: test.events index: _id_ dup key: { : { key: ObjectId('56c6dc1eff2e140920771ee5'), seq: 12 } }"
    }
  ]
}
```
Has we can see, the corresponding write result will be "`ok`" despite `writeErrors` being non-empty. Hence we fall in the `getOrElse` clause resulting in errors being mangled into a `GenericDriverException`.

jbgi:
While doing ordered bulk inserts I noticed that an error on a individual document get lost into a `GenericDriverException`.
I'd rather have a more structured `LastError`. (because I need to recover from duplicate key errors)

The reason of the behavior is that in `reactivemongo.api.collections.GenericCollection.Mongo26WriteCommand#send` we have:
```scala
       case Some(wr) if (wr.inError || (wr.hasErrors && ordered)) => {
          Future.failed(WriteResult.lastError(wr).
            getOrElse[Exception](GenericDriverException(
              s"write failure: $wr"
            )))
        }
```
and in `reactivemongo.api.commands.WriteResult#lastError`:
```scala
  def lastError(result: WriteResult): Option[LastError] = result match {
    case error @ LastError(_, _, _, _, _, _, _, _, _, _, _, _, _, _) => Some(error)
    case _ if (result.ok) => None
    case _ => Some(LastError(
      false, // ok
      result.errmsg,
```
Eg, if mongo (tested with 3.2) fail to insert a document due to a duplicate key, it will return an response like:
```
{
  ok: BSONInteger(1),
  n: BSONInteger(0),
  writeErrors: [
    0: {
      index: BSONInteger(0),
      code: BSONInteger(11000),
      errmsg: "E11000 duplicate key error collection: test.events index: _id_ dup key: { : { key: ObjectId('56c6dc1eff2e140920771ee5'), seq: 12 } }"
    }
  ]
}
```

Has we can see, the corresponding write result will be "`ok`" despite `writeErrors` being non-empty. Hence we fall in the `getOrElse` clause resulting in errors being mangled into a `GenericDriverException`.

### DIFF
--- a/driver/src/main/scala/api/commands/writeresult.scala
+++ b/driver/src/main/scala/api/commands/writeresult.scala
@@ -32,8 +32,7 @@ sealed trait WriteResult {
 object WriteResult {
   def lastError(result: WriteResult): Option[LastError] = result match {
     case error @ LastError(_, _, _, _, _, _, _, _, _, _, _, _, _, _) => Some(error)
-    case _ if (result.ok) => None
-    case _ => Some(LastError(
+    case _ if result.inError || result.hasErrors => Some(LastError(
       false, // ok
       result.errmsg,
       result.code,
@@ -49,6 +48,7 @@ object WriteResult {
       result.writeErrors,
       result.writeConcernError
     ))
+    case _ => None
   }
 
   /**


driver/src/main/scala/api/commands/writeresult.scala
```diff
@@ -32,8 +32,7 @@ sealed trait WriteResult {
 object WriteResult {
   def lastError(result: WriteResult): Option[LastError] = result match {
```
cchantep: Considering `lastError` this change is ok.

Considering the bulk use case mentioned, it should be the only case where a write op is `ok` but with `writeErrors`.

Additionally, I don't think this change is sufficient to have all the bulk errors in the corresponding `MultiBulkWriteResult`.

jbgi: yes, with this change, I get a failed future with a `LastError` instead of `GenericDriverException`.

To get a `MultiBulkWriteResult`(in a successful future) one have to use `ordered = false`.
I think this was intended (as per [comment](https://github.com/ReactiveMongo/ReactiveMongo/blob/13a2c9dcb5481f3c8af53ec9df4d8d52f16ba29d/driver/src/main/scala/api/collections/genericcollection.scala#L368)), but correct me if I'm wrong.
I have observed that the (correct) behavior for `ordered = false` was not changed by this PR.

cchantep: The comment corresponds to the `insert` use case, not bulk.

Would be nice to ensure there is a test checking all the error details in case of bulk.

Then the question is whether it should really fails, whereas MongoDB indicates `ok`.

jbgi: > the question is whether it should really fails, whereas MongoDB indicates ok.

If we decide that the Future should not fail then it would be a breaking change in semantic I guess. Could breaks applications that don't notice the change.

I will try to add a test to check failure behavior of bulk inserts. Could take me a bit of time to figure-out how/where so I'd appreciate any pointer.

cchantep: > If we decide that the Future should not fail then it would be a breaking change in semantic I guess. Could breaks applications that don't notice the change.

Actually it won't break the current API semantic, as currently a `ok` with some write errors happens as a `Future.successful` (but the current result inside may not have all details about the related errors).

jbgi: > Actually it won't break the current API semantic, as currently a ok with some write errors happens as a Future.successful

Only in the `ordered = false` case. On `ordered = true`we get a failed future with a `GenericDriverException` (or a `LastError` with this PR)

cchantep: I think to really fix the issue with bulk, some specific error handling should be done.
The previous or the new behaviour of `.lastError` is ok for individual write op.

jbgi: I would argue that even for individual write the api could be improved, ie. it return a `Future[WriteResult]` and `WriteResult` can contains errors. But in reality if the future is successful then the `WriteResult`  does not contains errors, because in that case a failed Future is used.
This lead to hesitation or duplication: the user does not really know from the type if it need to inspect the  `WriteResult` or the `Future`'s exception when looking for standard mongodb error codes.

So I would prefer we either always use `WriteResult`/`MultiBulkWriteResult` in case of mongo errors,
OR we use failed futures but in this case the `WriteResult` should no contains any errors.

cchantep: Even if there could be doubt about `WriteResult`, this case is documented and doesn't loose error detail.

On the other side, bulk is loosing such detail, and is the only case when there could be error even if `ok`.

jbgi: Not sure there we lose any detail for bulk writes:

- With `ordered=true` mongo stop at first error, so using `LastError` (as in this PR) in a failed future seams ok since there is at most on error.
- With `ordered=false` a successful future of `MultiBulkWriteResult` is return in all cases and (should?) contains all errors.

PR #628, opened 2017-02-06
Fix scaladoc of Macros.Options.SaveSimpleName

jbgi:
documentation fix: `SaveSimpleName` is for NOT using the fully-qualified class name.


cchantep:
Thx @jbgi

PR #633, opened 2017-02-23
BSON Support for Java8 Time

cchantep:


cchantep:
See https://github.com/ReactiveMongo/ReactiveMongo-BSON/pull/22

PR #642, opened 2017-04-08
Update Akka upper version to 2.4.17

cchantep:


driver/src/main/scala/core/actors.scala
```diff
@@ -983,6 +872,141 @@ trait MongoDBSystem extends Actor {
   override lazy val receive: Receive =
     processing.orElse(authReceive).orElse(fallback)
 
+  // ---
+
+  private def onIsMaster(response: Response) {
```
cchantep: Extract `IsMaster` handling from `receive`

PR #644, opened 2017-04-10
add aggregation pipeline stage AddFields (new in mongodb 3.4)

ornicar:


cchantep:
0.12.2-SNAPSHOT published again with it

Issue #654, opened 2017-04-25
Akka 2.5.0 breaking change


yannick-cw:
### ReactiveMongo Version (0.10.5, 0.11.6 / etc)
0.12.2

### Library Dependencies
akka `2.5.0`

### Expected Behavior
Reactive mongo should run like expected when I use akka `2.5.0` in my application.

### Actual Behavior
`driver.system.shutdown()`[line](https://github.com/ReactiveMongo/ReactiveMongo/blob/1136fadec66e97a413665788eb323b9a3ba32cb2/driver/src/main/scala/api/api.scala#L1142) and `.isTerminated` are not support anymore in akka `2.5.0`

Leads to 
```
java.lang.NoSuchMethodError: akka.actor.ActorSystem.shutdown()V
 	at reactivemongo.api.MongoDriver$SupervisorActor.postStop(api.scala:1142)
```
and
```
java.lang.NoSuchMethodError: akka.actor.ActorSystem.isTerminated()Z
 	at reactivemongo.api.MongoDriver.close(api.scala:893)
```

### Solution
update minimum support akka version and replace `shutdown` with `terminate` and `isTerminated` with `whenTerminated`

cchantep:
Hi,

As indicated the [documentation](http://reactivemongo.org/releases/0.12/documentation/release-details.html), ReactiveMongo 0.12.x is tested from Akka 2.3 to 2.4, not with 2.5 only released this month.

Breaking this backward compatibility is not planned.

> Please first ask on the [MailingList](https://groups.google.com/forum/m/?fromgroups#!forum/reactivemongo).

PR #655, opened 2017-04-27
Akka 2.5 compatibility

cchantep:
About #654


driver/src/main/scala/api/collections/genericcollection.scala
```diff
@@ -1019,4 +955,80 @@ trait GenericCollection[P <: SerializationPack with Singleton] extends Collectio
       ()
     }
   }
+
+  // ---
+
+  final class AggregatorContext[T](
```
cchantep: Split to allow to proper encoding of the type dependencies (between `Cursor`, `CursorProducer` and `CursorFlattener`)

driver/src/main/scala/api/collections/genericcollection.scala
```diff
@@ -1019,4 +955,80 @@ trait GenericCollection[P <: SerializationPack with Singleton] extends Collectio
...
+                new ChannelBufferWritableBuffer()
+              )(pack.writeToBuffer).toReadableBuffer).buffer
+
+            def resp = Response(
+              response.header,
+              Reply(0, resultCursor.cursorId, 0, firstBatch.size),
+              docs, response.info
+            )
+
+            cp.produce(DefaultCursor.getMore[P, T](pack, resp,
```
cchantep: Produce on the inner `Cursor` (providing the required `CursorOps`)

driver/src/main/scala/api/collections/genericcollection.scala
cchantep: cc @guyonvarch

PR #448, opened 2016-01-28
Add maxSize to collection stats results

srottenberg:
The `maxSize` information was missing from the stats command results.
It can be useful for capped collections, for example to calculate the remaining available space in the collection.

PR #674, opened 2017-06-23
Changing type of index size values for stats command to Double.

ehoner:
# Pull Request Checklist

## Fixes

Index sizes can be larger than an `Int` can support. The documentation in mongo does not make clear the maximum value, only that it is a "number". 

https://docs.mongodb.com/manual/reference/command/collStats/#example

https://docs.mongodb.com/manual/reference/command/collStats/#collStats.totalIndexSize

## Purpose

This changes the types to `Double` to correspond to the other size values returned by the command.

## Background Context

This bug was discovered after adding several indices to an existing collection. The `stats()` command began returning this error:
```reactivemongo.bson.BSONDouble cannot be cast to reactivemongo.bson.BSONInteger : Error```

These were the values returned by `stats()`:
```javascript
{
                "sharded" : false,
                "primary" : "xxxx",
                "ns" : "yyyy",
                "count" : 2767914,
                "size" : 2517842272,
                "avgObjSize" : 909,
                "numExtents" : 22,
                "storageSize" : 3918798848,
                "lastExtentSize" : 1021497344,
                "paddingFactor" : 1,
                "paddingFactorNote" : "paddingFactor is unused and unmaintained in 3.0. It remains hard coded to 1.0 for compatibility only.",
                "userFlags" : 1,
                "capped" : false,
                "nindexes" : 5,
                "totalIndexSize" : 2659693680,
                "indexSizes" : {
                                "_id_" : 97400688,
                                "secureDistinctKey_1_secureId_1_observationDate_-1_updateDate_-1" : 1146806640,
                                "secureId_1_observationDate_-1_updateDate_-1" : 509430208,
                                "secureId_1_userId_1" : 612112592,
                                "userId_1_observationDate_-1_createDate_-1" : 293943552
                },
                "ok" : 1
}
```
The `totalIndexSize` was larger than an `Int` can support. I presume the same could occur for any single index as well.

## References

ReactiveMongo/ReactiveMongo#448 added `maxSize` and also matched type to `Double`

### DIFF
--- a/driver/src/main/scala/api/commands/bson/instanceadministration.scala
+++ b/driver/src/main/scala/api/commands/bson/instanceadministration.scala
@@ -131,10 +131,10 @@ object BSONCollStatsImplicits {
       doc.getAs[BSONNumberLike]("paddingFactor").map(_.toDouble),
       doc.getAs[BSONNumberLike]("systemFlags").map(_.toInt),
       doc.getAs[BSONNumberLike]("userFlags").map(_.toInt),
-      doc.getAs[BSONNumberLike]("totalIndexSize").map(_.toDouble).get,
+      doc.getAs[Int]("totalIndexSize").getOrElse(-1),
       {
         val indexSizes = doc.getAs[BSONDocument]("indexSizes").get
-        (for (kv <- indexSizes.elements) yield kv.name -> kv.value.asInstanceOf[BSONDouble].value).toList
+        (for (kv <- indexSizes.elements) yield kv.name -> indexSizes.getAs[Int](kv.name).getOrElse(-1)).toList
       },
       doc.getAs[BSONBooleanLike]("capped").fold(false)(_.toBoolean),
       doc.getAs[BSONNumberLike]("max").map(_.toLong),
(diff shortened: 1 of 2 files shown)


driver/src/main/scala/api/commands/instanceadministration.scala
```diff
@@ -70,13 +71,13 @@ case class CollStatsResult(
     paddingFactor: Option[Double],
     systemFlags: Option[Int],
     userFlags: Option[Int],
-    totalIndexSize: Int,
-    sizePerIndex: List[(String, Int)],
+    totalIndexSize: Double,
```
cchantep: That's breaking backward compat

driver/src/main/scala/api/commands/bson/instanceadministration.scala
```diff
@@ -131,10 +131,10 @@ object BSONCollStatsImplicits {
       doc.getAs[BSONNumberLike]("paddingFactor").map(_.toDouble),
       doc.getAs[BSONNumberLike]("systemFlags").map(_.toInt),
       doc.getAs[BSONNumberLike]("userFlags").map(_.toInt),
-      doc.getAs[BSONNumberLike]("totalIndexSize").map(_.toInt).get,
+      doc.getAs[Int]("totalIndexSize").getOrElse(-1),
```
cchantep: Strictly reading as `Int` is not recommanded with MongoDB commands, which are used to change the type of the number properties from one release to another

cchantep: I doubt the change worth it

ehoner:
Thanks for your review.

PR #693, opened 2017-08-25
Add asynchronous APIs to open connection and to close driver.

jbgi:


driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -2,26 +2,19 @@ package reactivemongo.api
 
 import java.util.concurrent.atomic.AtomicLong
 
-import scala.util.{ Try, Failure, Success }
```
cchantep: Please keep the import formatting, including the explicit imports rather than `_`

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -98,17 +91,24 @@ class MongoDriver(
...
+    Await.result(askClose()(timeout), timeout) // Unsafe
+  }
 
-    // Terminate actors used by MongoConnections
-    connections.foreach(_.close())
+  def askClose()(implicit timeout: FiniteDuration): Future[Unit] = {
+    logger.info(s"[$supervisorName] Closing instance of ReactiveMongo driver")
 
+    val p = Promise[Unit]()
+    import ExecutionContext.Implicits.global
```
cchantep: Rather have an implicit parameter `ExecutionContext`

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -98,17 +91,24 @@ class MongoDriver(
...
     // Tell the supervisor to close.
     // It will shut down all the connections and monitors
-    // and then shut down the ActorSystem as it is exiting.
-    supervisorActor ! Close
-
-    Await.result(systemClose(Some(timeout)), timeout) // Unsafe
+    (supervisorActor ? Close)(Timeout(timeout)).onComplete { sResult =>
```
cchantep: I don't see need to use `Promise` there, better to compose from `?` result

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -136,6 +136,20 @@ class MongoDriver(
...
+   * Creates a new MongoConnection.
+   *
+   * $seeConnectDBTutorial
+   *
+   * @param nodes $nodesParam
+   * @param authentications $authParam
+   * @param name $connectionNameParam
+   * @param options $optionsParam
+   */
+  def askConnection(nodes: Seq[String], options: MongoConnectionOptions = MongoConnectionOptions(), authentications: Seq[Authenticate] = Seq.empty, name: Option[String] = None): Future[MongoConnection] = {
```
cchantep: I'm not 100% by the `askConnection` naming. It depends whether the former connection would be deprecated in favor of such one, what is not yet decided.

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -200,6 +233,17 @@ class MongoDriver(
    *
    * $seeConnectDBTutorial
    *
+   * @param uri the URI to be parsed by [[reactivemongo.api.MongoConnection.parseURI]]
+   * @param name $connectionNameParam
+   * @param strictUri $strictUriParam
+   */
+  def askConnection(uri: String, name: Option[String], strictUri: Boolean): Future[MongoConnection] = Future.fromTry(MongoConnection.parseURI(uri)).flatMap(askConnection(_, name, strictUri))(system.dispatcher)
```
cchantep: Such connection not using strict URI should rather be kept `private` (or be inlined where needed), as at some point accepting URI with invalid option will no longer be accepted (no such not be part of new API)

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -136,6 +136,20 @@ class MongoDriver(
    * @param options $optionsParam
    */
   def connection(nodes: Seq[String], options: MongoConnectionOptions = MongoConnectionOptions(), authentications: Seq[Authenticate] = Seq.empty, name: Option[String] = None): MongoConnection = {
+    Await.result(askConnection(nodes, options, authentications, name), Duration.Inf)
+  }
+
+  /**
+   * Creates a new MongoConnection.
+   *
+   * $seeConnectDBTutorial
```
cchantep: As far as I see undefined?

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -227,13 +271,28 @@ class MongoDriver(
    * @param strictUri $strictUriParam
    */
   def connection(parsedURI: MongoConnection.ParsedURI, name: Option[String], strictUri: Boolean): Try[MongoConnection] = {
+    val connection = askConnection(parsedURI, name, strictUri)
+    Await.ready(connection, Duration.Inf)
+    connection.value.get
```
cchantep: Something wrong there

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -301,22 +378,20 @@ class MongoDriver(
...
+          context.stop(self)
+          sender() ! Closed
+        } else {
+          context.become(closing(mutable.Queue[ActorRef](sender)))
+          driver.connectionMonitors.values.foreach(_.close())
+        }
       }
     }
 
-    @deprecated("Will be private", "0.12.6")
```
cchantep: At least this should be kept deprecated, as it in anyway it should be part of the API. If there is still need for it would be renamed as a `private` function.

cchantep: Btw as this function won't be removed until next major release, this change is breaking compatibility. Have a look at MiMa reports.

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -301,22 +378,20 @@ class MongoDriver(
...
       case Close => {
         logger.debug(s"[$supervisorName] Close the supervisor")
 
-        if (isEmpty) context.stop(self)
-        else context.become(closing(Duration.Zero))
+        if (isEmpty) {
+          context.stop(self)
+          sender() ! Closed
+        } else {
+          context.become(closing(mutable.Queue[ActorRef](sender)))
```
cchantep: I don't see the point for the function (see next comment about `closing`) creating the new `Receive` context to be passed the mutable queue.

As using such mutable queue is internal implementation details for such function, this one should be responsible for initializing it (maybe as `@volatile`).

driver/src/main/scala/api/MongoDriver.scala
jbgi: not sure why this was still deprecated with "Will be private", as the whole `SupervisorActor` class is already private anyway... but I will have a look a MiMa reports.

cchantep: Right, compatibility should not be an issue there.

driver/src/main/scala/api/MongoDriver.scala
jbgi: I think I fixed all other review points. Let me known when this point is decided.

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -98,17 +100,18 @@ class MongoDriver(
    * Awaits the termination until the timeout is expired.
    */
```
cchantep: For this one adding the `@deprecated` already seems right

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -198,6 +210,17 @@ class MongoDriver(
   /**
    * Creates a new MongoConnection from URI.
    *
+   * @param uri the URI to be parsed by [[reactivemongo.api.MongoConnection.parseURI]]
+   * @param name $connectionNameParam
+   */
+  def askConnection(uri: String, name: Option[String]): Future[MongoConnection] = MongoConnection.parseURI(uri) match {
```
cchantep: `uriStrict` ?

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -301,40 +345,42 @@ class MongoDriver(
...
+          driver.connectionMonitors.values.foreach(_.close())
+        }
       }
     }
 
-    @deprecated("Will be private", "0.12.6")
-    def closing(shutdownTimeout: FiniteDuration): Receive = {
-      case AddConnection(name, _, _, _) =>
-        logger.warn(s"[$supervisorName] Refusing to add connection while MongoDriver is closing: $name")
+    def closing(): Receive = {
```
cchantep: `def closing: Receive`

driver/src/main/scala/api/MongoDriver.scala
jbgi: done!

driver/src/main/scala/api/MongoDriver.scala
jbgi: yeah, uri is now always strict for new askConnection API, so maybe we should simply use `uri` as param name everywhere in the new api? or do you prefer we use `uriStrict`?

driver/src/main/scala/api/MongoDriver.scala
jbgi: done

cchantep:
Scalariform issue : https://travis-ci.org/ReactiveMongo/ReactiveMongo/jobs/269266862#L402

driver/src/main/scala/api/MongoDriver.scala
cchantep: https://github.com/cchantep/ReactiveMongo/commit/2c62e4ec4cd69929a2f887012865bb5478c43ebc

jbgi: 👍

cchantep: You can cherry pick the API changes so this PR can go on

driver/src/main/scala/api/Driver.scala
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  private case class AddConnection(
+    name: String,
+    nodes: Seq[String],
+    options: MongoConnectionOptions,
+    mongosystem: ActorRef)
+  
+  // For testing only
+  private[api] def addConnectionMsg(
```
cchantep: @jbgi Need to update [`src/test/scala/ApiTest.scala`](https://github.com/jbgi/ReactiveMongo/blob/d3bd78caa90a8d6651715e4b09bad02ab5120bc0/driver/src/test/scala/ApiTesting.scala#L35) using that

PR #695, opened 2017-08-26
use scala.collection.breakOut to speed up the MapReader

ornicar:
Same behaviour, but skips a `Map` creation.

See https://docs.scala-lang.org/tutorials/FAQ/breakout.html


cchantep:
Thx

PR #698, opened 2017-09-06
Do not silently restart a new tailable cursor after maxDocs is reached.

jbgi:
Should resolve failing test of https://github.com/ReactiveMongo/ReactiveMongo-Streaming/pull/56.

### DIFF
--- a/driver/src/main/scala/api/cursor.scala
+++ b/driver/src/main/scala/api/cursor.scala
@@ -552,9 +552,8 @@ object DefaultCursor {
               Future.failed[Option[Response]](err)
           }
         } else {
-          logger.debug("[tailResponse] Current cursor exhausted, renewing...")
-          DelayedFuture(500, connection.actorSystem).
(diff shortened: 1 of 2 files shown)


cchantep:
Any such change would need to update some cursor tests or to add some to demonstrate the validity of the change and the non-regression.

jbgi:
I will try to come-up with some tests. But since this change impact only an api exclusively used by ReactiveMongo-Streaming (currently not tested in driver project), validating non regression would require running tests of that project against this PR. Is that possible?

cchantep:
Tailable cursor is not only used by streaming. If there is an issue only in streaming, then no change there should fix it.

jbgi:
Right, the method is used, it was just intellij "find usage" failing me.

jbgi:
Added a test that ([fails](https://travis-ci.org/ReactiveMongo/ReactiveMongo/jobs/273031554#L2458)) without this fix.

PR #722, opened 2017-11-20
Add support to use x.509 certificates to authenticate clients

konradwudkowski:
# Pull Request Checklist

## Purpose

This PR adds support to use x.509 certificates to authenticate clients as explained in more details here: https://docs.mongodb.com/manual/tutorial/configure-x509-client-authentication

## References

This PR builds on top of [606](https://github.com/ReactiveMongo/ReactiveMongo/pull/606) and is inspired by and meant to supersede [617](https://github.com/ReactiveMongo/ReactiveMongo/pull/617)


konradwudkowski:
This is a continuation of work mentioned here: https://groups.google.com/forum/#!topic/reactivemongo/WSuyC10VxwM

This is what me and @githubmo did

.ci_scripts/runIntegration.sh
```diff
@@ -14,10 +14,16 @@ else
     SBT_ARGS="$SBT_ARGS -Dtest.authMode=cr"
 fi
 
-if [ "$MONGO_PROFILE" = "invalid-ssl" -o "$MONGO_PROFILE" = "mutual-ssl" ]; then
+if [ "$MONGO_PROFILE" != "x509" ]; then
+    TEST_OPTS="$TEST_OPTS,x509"
+else
+    TEST_CLASSES=X509DriverSpec
```
cchantep: In case of X509, not only that tests must be executed, but all the integration tests to cover there is no regression/side-effect due to that auth.

.ci_scripts/unitTests.sh
```diff
@@ -1,9 +1,9 @@
 #! /bin/bash
 
-if [ `git grep localhost | grep 'src/test' | grep -vi 'Common.scala' | wc -l` -ne 0 ]; then
+if [ `git grep localhost | grep 'src/test' | grep -vi 'Common.scala' | grep -vi 'X509DriverSpec.scala' | wc -l` -ne 0 ]; then
```
cchantep: It's either integration tested or unit tested, but I doubt this one is both

driver/src/main/scala/api/MongoConnection.scala
```diff
@@ -412,27 +412,48 @@ object MongoConnection {
...
 
+          case WithAuthNoPassword(user, hostsPortsAndDbName) if options.authMode == X509Authentication => {
+            val (db, hosts) = parseHostsAndDbName(hostsPortsAndDbName)
+
+            db.fold[ParsedURI](throw new URIParsingException(s"Could not parse URI '$uri': authentication information found but no database name in URI")) { database =>
+
+              options.authMode match {
+                case _ => ParsedURI(hosts, options, unsupportedKeys, Some(database),
+                  Some(Authenticate(options.authSource.
+                    getOrElse(database), user, "")))
```
cchantep: `""` seems weird to me

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -1,16 +1,15 @@
 package reactivemongo.api
 
-import scala.util.{ Try, Failure, Success }
-
+import scala.util.{ Failure, Success, Try }
```
cchantep: Please gather the import groups as before

driver/src/main/scala/core/actors/MongoDBSystem.scala
```diff
@@ -1351,6 +1351,20 @@ final class StandardDBSystem private[reactivemongo] (
...
+  val supervisor: String,
+  val name: String,
+  val seeds: Seq[String],
+  val initialAuthenticates: Seq[Authenticate],
+  val options: MongoConnectionOptions) extends MongoDBSystem with MongoX509Authentication {
+
+  def newChannelFactory(effect: Unit): ChannelFactory =
+    new ChannelFactory(supervisor, name, options)
+
+  @deprecated("Initialize with an explicit supervisor and connection names", "0.11.14")
```
cchantep: ?

driver/src/main/scala/core/commands/authentication.scala
```diff
@@ -1,13 +1,7 @@
 package reactivemongo.core.commands
 
-import reactivemongo.bson.{
```
cchantep: Please keep the import formatting

driver/src/test/scala/MongoURISpec.scala
```diff
@@ -1,9 +1,5 @@
-import reactivemongo.api.{
```
cchantep: Please keep the import formatting

driver/src/test/scala/X509DriverSpec.scala
```diff
@@ -0,0 +1,75 @@
+
+import org.specs2.concurrent.ExecutionEnv
+import reactivemongo.api.{ DefaultDB, X509Authentication }
+import reactivemongo.bson.BSONDocument
+import reactivemongo.core.actors.Exceptions.{ InternalState, PrimaryUnavailableException }
+import reactivemongo.core.nodeset.Authenticate
+import scala.concurrent.Await
+
+class X509DriverSpec(implicit ee: ExecutionEnv) extends org.specs2.mutable.Specification {
```
cchantep: Would rather execute with the tests existing in `DriverSpec`, with appropriate tags, maybe with a trait mixin

driver/src/main/scala/api/MongoConnection.scala
```diff
@@ -412,27 +412,48 @@ object MongoConnection {
...
 
-        ParsedURI(hosts, options, unsupportedKeys, db, None)
+        options.authMode match {
+          case X509Authentication => ParsedURI(hosts, options, unsupportedKeys, db,
+            Some(Authenticate(db.getOrElse(throw new URIParsingException(s"Could not parse URI '$uri': authentication information found but no database name in URI")), "", "")))
+          case _ => ParsedURI(hosts, options, unsupportedKeys, db, None)
+        }
       } else {
         val WithAuth = """([^:]+):([^@]*)@(.+)""".r
+        val WithAuthNoPassword = """([^:]+)@(.+)""".r
```
cchantep: For me doesn't match the URL scheme with `mongodb://`

konradwudkowski: @cchantep do you mean that password should be supplied even if not required? For x509 user is needed but private key is used instead of a password effectively

driver/src/main/scala/api/MongoConnection.scala
konradwudkowski: @cchantep there is no password, could you please suggest a better alternative

konradwudkowski:
hi @cchantep, thanks a lot for your feedback. I've made a few changes. Could you please review again. Thanks!

driver/src/main/scala/api/MongoConnection.scala
cchantep:
```scala
val WithAuth = """([^:]+)(|:[^@]*)@(.+)""".r

"user@foo/db" match { case WithAuth(a, b, c) => println(s"OK: '$a' '${b drop 1}' '$c'"); case _ => println("KO") }
// --> OK: 'user' '' 'foo/db'

"user:pwd@foo/db" match { case WithAuth(a, b, c) => println(s"OK: '$a' '${b drop 1}' '$c'"); case _ => println("KO") }
// --> OK: 'user' 'pwd' 'foo/db'
```

driver/src/test/scala/Common.scala
```diff
@@ -167,4 +166,45 @@ object Common {
...
+    ifX509(Seq(Authenticate("", certSubject, "")))(otherwise = Nil)
+
+  def certSubject: String =
+    Option(System.getProperty("test.clientCertSubject"))
+      .getOrElse(sys.error("Client cert subject required if X509 auth enabled"))
+
+  def increaseTimeoutIfX509(timeout: FiniteDuration): FiniteDuration =
+    ifX509(timeout * 10)(otherwise = timeout)
+
+  def makeConnection(nodes: Seq[String], options: MongoConnectionOptions): MongoConnection = {
```
cchantep: I don't see benefit for those `makeConnectionX` which are essentially oneliner aliases

driver/src/test/scala/Common.scala
```diff
@@ -167,4 +166,45 @@ object Common {
...
+    }
+
+  def defaultAuthentications: Seq[Authenticate] =
+    ifX509(Seq(Authenticate("", certSubject, "")))(otherwise = Nil)
+
+  def certSubject: String =
+    Option(System.getProperty("test.clientCertSubject"))
+      .getOrElse(sys.error("Client cert subject required if X509 auth enabled"))
+
+  def increaseTimeoutIfX509(timeout: FiniteDuration): FiniteDuration =
```
cchantep: Why?

driver/src/test/scala/Common.scala
```diff
@@ -167,4 +166,45 @@ object Common {
 
     slowProxy.stop()
   }
+
+}
+
+trait CommonAuth {
```
cchantep: I would rather keep in `Common`

driver/src/main/scala/core/nodeset/authentication.scala
```diff
@@ -57,4 +60,8 @@ case class ScramSha1Authenticating(
     s"Authenticating($db, $user})"
 }
 
+case class X509Authenticating(db: String, user: String, password: String) extends Authenticating {
```
cchantep: Having an empty string instead of nothing (or at least `None`) is quite weird. Will need changes.

driver/src/main/scala/core/commands/authentication.scala
```diff
@@ -339,6 +339,31 @@ object CrAuthenticate extends BSONCommandResultMaker[SuccessfulAuthentication] {
   }
 }
 
+private[core] case class X509Authenticate(user: String) extends Command[SuccessfulAuthentication] {
```
cchantep: Rather `Option[String]`

driver/src/test/scala/Common.scala
konradwudkowski: Tests got very flaky until I've increased the timeout. Running individual classes works but running all it tests was flaky

driver/src/test/scala/Common.scala
konradwudkowski: There are 2 `Common` objects, one for most tests and one for the jmx stuff and both required the same handling to enable X509 client auth.

.travis.yml
```diff
@@ -32,7 +32,8 @@ env:
   - CI_CATEGORY=INTEGRATION_TESTS AKKA_VERSION=2.5.6 ITERATEES_VERSION=2.6.1 MONGO_PROFILE=invalid-ssl MONGO_VER=3_4
   - CI_CATEGORY=INTEGRATION_TESTS AKKA_VERSION=2.5.6 ITERATEES_VERSION=2.6.1 MONGO_PROFILE=mutual-ssl MONGO_VER=3_4
   - CI_CATEGORY=INTEGRATION_TESTS AKKA_VERSION=2.5.6 ITERATEES_VERSION=2.6.1 MONGO_PROFILE=rs MONGO_VER=3_4
-matrix:
```
cchantep: Need to update the matrix generator in the build

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -4,13 +4,13 @@ import scala.util.{ Try, Failure, Success }
...
 import scala.concurrent.{ Await, ExecutionContext }
 import scala.concurrent.duration.{ Duration, FiniteDuration, SECONDS }
-
 import com.typesafe.config.Config
-
-import akka.actor.ActorRef
-
+import akka.actor.{ ActorRef, Props }
+import akka.util.Timeout
+import reactivemongo.core.actors.{ LegacyDBSystem, MongoDBSystem, StandardDBSystem, StandardDBSystemWithX509 }
```
cchantep: Unused imports, remove rather than reformat

driver/src/main/scala/api/Driver.scala
```diff
@@ -1,26 +1,15 @@
...
 import scala.concurrent.{ ExecutionContext, Future }
 import scala.concurrent.duration.{ FiniteDuration, SECONDS }
-
 import com.typesafe.config.Config
-
 import akka.util.Timeout
 import akka.actor.{ Actor, ActorRef, ActorSystem, Props, Terminated }
 import akka.pattern.ask
-
-import reactivemongo.core.actors.{
```
cchantep: Please keep explicit imports

driver/src/main/scala/api/Driver.scala
```diff
@@ -1,26 +1,15 @@
 package reactivemongo.api
 
 import java.util.concurrent.atomic.AtomicLong
-
```
cchantep: Please keep import group/spacing

driver/src/test/scala/Common.scala
konradwudkowski: Renamed to overloaded makeConnetion in both cases. The reasons for having both is that some tests create a separate instance of a driver and control its lifecycle

.travis.yml
konradwudkowski: Could you please tell me what needs updating there? I've added a new profile for `x509`

driver/src/main/scala/api/MongoDriver.scala
konradwudkowski: :+1:

driver/src/main/scala/api/Driver.scala
konradwudkowski: :+1:

driver/src/main/scala/api/Driver.scala
konradwudkowski: :+1:

driver/src/main/scala/api/MongoConnection.scala
konradwudkowski: I'm not sure your `password.drop(1)` is immediately understandable but I've made a similar change that uses your regex

driver/src/test/scala/Common.scala
cchantep: The test classes are separate

konradwudkowski: In order to run both standard and jmx tests with and without x509 auth I need to modify the connection. Logic to do it is similar so I put it in a separate trait that I use from both places. The trait has only auth related functionality. Can you please suggest a better way?

.travis.yml
cchantep: This matrix is/should be generated using the task in the SBT build

.ci_scripts/addX509User.sh
```diff
@@ -0,0 +1,40 @@
...
+SCRIPT_DIR="$3"
+ENV_FILE="$4"
+
+PORT=27018
+
+mongod --dbpath $DB_PATH --port $PORT --fork --logpath /tmp/mongod.log
+
+CLIENT_CERT_SUBJECT=`openssl x509 -in $SCRIPT_DIR/client.pem -inform PEM -subject -nameopt RFC2253 | grep subject | awk '{sub("subject= ",""); print}'`
+
+cat > /tmp/add_subject_as_user.js <<EOF
```
cchantep: `mktemp`

.ci_scripts/runIntegration.sh
```diff
@@ -14,10 +14,17 @@ else
     SBT_ARGS="$SBT_ARGS -Dtest.authMode=cr"
 fi
 
-if [ "$MONGO_PROFILE" = "invalid-ssl" -o "$MONGO_PROFILE" = "mutual-ssl" ]; then
+if [ "$MONGO_PROFILE" != "x509" ]; then
+    TEST_OPTS="$TEST_OPTS,x509"
```
cchantep: `# add to excludes`

.ci_scripts/runIntegration.sh
```diff
@@ -14,10 +14,17 @@ else
...
+else
+    TEST_OPTS="$TEST_OPTS,scram_auth,cr_auth"
```
cchantep: `# add to excludes`

.ci_scripts/setupEnv.sh
```diff
@@ -60,17 +60,25 @@ EOF
         cat >> "$MONGO_CONF" << EOF
     allowInvalidCertificates: true
 EOF
-    else
+    elif [ "$MONGO_PROFILE" = "mutual-ssl" -o "$MONGO_PROFILE" = "x509" ]; then
```
cchantep: ?

.ci_scripts/setupEnv.sh
```diff
@@ -60,17 +60,25 @@ EOF
...
         # mutual-ssl
         cat >> "$MONGO_CONF" << EOF
     CAFile: $SCRIPT_DIR/client.pem
 EOF
 
-        keytool -importkeystore  -srcstoretype PKCS12 \
-            -srckeystore "$SCRIPT_DIR/keystore.p12" \
-            -destkeystore /tmp/keystore.jks \
-            -storepass $SSL_PASS -srcstorepass $SSL_PASS
+    keytool -importkeystore  -srcstoretype PKCS12 \
```
cchantep: Indent issue

.ci_scripts/addX509User.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+set -e -x
+
+echo
+echo "adding x509 Certificate subject as a User"
+# https://docs.mongodb.com/manual/tutorial/configure-x509-client-authentication/#add-x-509-certificate-subject-as-a-user
+
+PATH="$1"
```
cchantep: `PATH` not need if this one is `source`d as "sub-script"

.ci_scripts/addX509User.sh
```diff
@@ -0,0 +1,40 @@
...
+DB_PATH="$2"
+SCRIPT_DIR="$3"
+ENV_FILE="$4"
+
+PORT=27018
```
cchantep: Would rather get it from `PRIMARY_HOST` available in `setupEnv.sh`

driver/src/main/scala/api/MongoConnection.scala
```diff
@@ -412,20 +412,31 @@ object MongoConnection {
 
       // ---
 
+      val (unsupportedKeys, options) = opts
```
cchantep: Blank line after the `val` block please

driver/src/main/scala/api/MongoConnection.scala
```diff
@@ -412,20 +412,31 @@ object MongoConnection {
...
+        options.authMode match {
+          case X509Authentication => ParsedURI(hosts, options, unsupportedKeys, db,
+            Some(Authenticate(db.getOrElse(throw new URIParsingException(s"Could not parse URI '$uri': authentication information found but no database name in URI")), "", "")))
+          case _ => ParsedURI(hosts, options, unsupportedKeys, db, None)
+        }
       } else {
-        val WithAuth = """([^:]+):([^@]*)@(.+)""".r
+
+        val WithAuth = """([^:]+)(|:[^@]*)@(.+)""".r
+        object PassExtractor {
```
cchantep: I don't think that's required: there will always be a matching group (as it's always `Some`), no extraction, so `drop(1)` is simpler and makes more sense.

driver/src/main/scala/api/MongoConnection.scala
```diff
@@ -412,20 +412,31 @@ object MongoConnection {
...
-        ParsedURI(hosts, options, unsupportedKeys, db, None)
+        options.authMode match {
+          case X509Authentication => ParsedURI(hosts, options, unsupportedKeys, db,
+            Some(Authenticate(db.getOrElse(throw new URIParsingException(s"Could not parse URI '$uri': authentication information found but no database name in URI")), "", "")))
+          case _ => ParsedURI(hosts, options, unsupportedKeys, db, None)
+        }
       } else {
-        val WithAuth = """([^:]+):([^@]*)@(.+)""".r
+
+        val WithAuth = """([^:]+)(|:[^@]*)@(.+)""".r
```
cchantep: Blank line after `val` rather than before there please

driver/src/main/scala/api/MongoDriver.scala
```diff
@@ -4,11 +4,8 @@ import scala.util.{ Try, Failure, Success }
 
 import scala.concurrent.{ Await, ExecutionContext }
 import scala.concurrent.duration.{ Duration, FiniteDuration, SECONDS }
-
```
cchantep: Please keep the formatting

driver/src/main/scala/core/MongoX509Authentication.scala
```diff
@@ -0,0 +1,43 @@
+package reactivemongo.core.actors
+
+import reactivemongo.bson.lowlevel.{ DoubleField, LowLevelBsonDocReader }
```
cchantep: This will be deprecated

driver/src/test/scala/Common.scala
cchantep: Anyway would directly call `driver.connection` rather than introducing such aliases

driver/src/test/scala/Common.scala
cchantep: If the suite runs fine locally, this extra timeout should be removed for now until a cause for it could be clearly identified

project/ReactiveMongo.scala
```diff
@@ -902,5 +902,5 @@ object Version {
       libraryDependencies ++= Seq(specs) ++ logApi,
       pomPostProcess := providedInternalDeps
     ).enablePlugins(CopyPasteDetector).
-    dependsOn(driver)
+    dependsOn(driver % "test->test;compile->compile")
```
cchantep: That's sharing all the tests classes for one `val`

.travis.yml
konradwudkowski: Updated to generate env/matrix with `test:travisEnv`

.ci_scripts/setupEnv.sh
konradwudkowski: fixed

.ci_scripts/setupEnv.sh
konradwudkowski: changed

.ci_scripts/runIntegration.sh
konradwudkowski: added comment

.ci_scripts/runIntegration.sh
konradwudkowski: added comment

.ci_scripts/addX509User.sh
konradwudkowski: :+1:

.ci_scripts/addX509User.sh
konradwudkowski: :+1:

.ci_scripts/addX509User.sh
konradwudkowski: :+1:

driver/src/test/scala/Common.scala
konradwudkowski: It's not about aliases here thought. Inside of `Common` you could, but many tests obtain connection not by using `connection` variable but instead use `driver.connection(...)` and modify some parameters. These tests would have to know about `MONGO_PROFILE` being potentially set to `x509` and explicitly add authentication. Here they work the same with or without `x509`

driver/src/main/scala/api/MongoConnection.scala
konradwudkowski: Regexes are by their nature not very readable and it's not immediately clear that password may start with a colon if you just use `pass.drop(1)`.  Also `pass` is used twice which means that a variable should be used rather than a duplication of `pass.drop(1)`. And I find `stripPrefix(":")` much better / more explicit than `drop(1)`. Having said that if you strongly prefer `drop(1)` please suggest variable names in the pattern match and later and I will refactor.

project/ReactiveMongo.scala
konradwudkowski: The only way I know to avoid it is to duplicate all the logic. Please suggest an alternative.

cchantep: Duplicate is not always evil ...

driver/src/main/scala/api/MongoConnection.scala
cchantep: 1. Using a custom extract when there is only one `Some` case doesn't make sense for me. Whatever is the normalization operation (`stripPrefix` or  `.drop(1)`), it should be done in the body of the `case WithAuth`.
2. In anyway, the regexp guarantee that either this part is already empty (nothing can be dropped), or it starts with `:`, having the `stripPrefix` checking it again is not useful.

.travis.yml
cchantep: thx

driver/src/test/scala/Common.scala
cchantep: For me it's aliasing, a also the call for such functions can be reverted back to `driver.connection`, the default connection in `Commons` being the only place where the test profile is known.

driver/src/main/scala/api/MongoConnection.scala
konradwudkowski: stripPrefix doesn't check, it strips the prefix, and provides ready to use pass with the colon removed

I don't want to argue over this really. I can do as below if you wish:

```

        val WithAuth = """([^:]+)(|:[^@]*)@(.+)""".r

        useful match {
          case WithAuth(user, pass, hostsPortsAndDbName) => {
            val (db, hosts) = parseHostsAndDbName(hostsPortsAndDbName)

            db.fold[ParsedURI](throw new URIParsingException(s"Could not parse URI '$uri': authentication information found but no database name in URI")) { database =>

              val passWithoutColon = pass.stripPrefix(":")
              
              if (options.authMode == X509Authentication && passWithoutColon.nonEmpty) {
                throw new URIParsingException("You should not provide a password when authenticating with X509 authentication")
              }

              ParsedURI(hosts, options, unsupportedKeys, Some(database),
                Some(Authenticate(options.authenticationDatabase.
                  getOrElse(database), user, passWithoutColon)))
            }
          }
```

cchantep: https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/immutable/StringLike.scala#L158

cchantep:
Fixed by #743

PR #725, opened 2017-12-05
Add BSONHandler able to process UUID

karol-szuster:
# Pull Request Checklist

## Purpose

This PR adds BSONHandler that handles java.util.UUID

### DIFF
--- a/bson/src/main/scala/DefaultBSONHandlers.scala
+++ b/bson/src/main/scala/DefaultBSONHandlers.scala
@@ -1,5 +1,8 @@
 package reactivemongo.bson
 
+import java.nio.ByteBuffer
+import java.util.UUID
+
 import scala.collection.generic.CanBuildFrom
 import scala.util.Try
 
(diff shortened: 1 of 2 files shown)


bson/src/main/scala/DefaultBSONHandlers.scala
```diff
@@ -75,6 +78,23 @@ trait DefaultBSONHandlers {
     def write(date: Date) = BSONDateTime(date.getTime)
   }
 
+  implicit object BSONUUID extends BSONHandler[BSONBinary, UUID] {
```
cchantep: Rather as `BSONString`

karol-szuster: My goal was to be compatible with `UUID` from mongodb. If you open mongodb shell and do something like this:

```
> new UUID("00112233445566778899AABBCCDDEEFF")
```

you will notice that mongo will return:

```
BinData(3,"ABEiM0RVZneImaq7zN3u/w==")
```

That's why it's implemented this way.

cchantep: I doubt that's the expected behaviour in most of the case

karol-szuster: I think it is expected behavior for numerous reasons:
* Acording to [bson specification](http://bsonspec.org/spec.html) there is dedicated binary subtype for storing UUID so I think it is more idiomatic to store it this way
* UUID are very often used to search in database and mongodb indexes are optimized to handle certain array sizes (source: [Optimize Query Performance](https://docs.mongodb.com/v3.4/tutorial/optimize-query-performance-with-indexes-and-projections/). I haven't made any tests but I believe that it should be faster than text search.
* Obviously it uses less storage this way.
* Other people are doing this for example guys from PyMongo please read [this stackoverflow question](https://stackoverflow.com/questions/12030709/why-does-pymongo-encode-uuid-uuid1-as-a-bsonbinary)

cchantep: I'm not convinced it should be the default behaviour. You can find many project storing UUID as string, as in many case the binary optimization doesn't offer much improvement.

cchantep: From the people I could have asked about, most are storing it as BSON string, so for now I tend to have that string codec as default, and the binary one available but not the implicit scope by default.

PR #730, opened 2018-01-15
Added updated samples

ricsirigu:
Added a sample app with the latest version of ReactiveMongo and play 2.6


ricsirigu:
Removed @cchantep

PR #748, opened 2018-06-11
Improve Delete api

jtjeferreira:
# Pull Request Checklist

* [X] Have you added tests for any changed functionality? Yes by changing existing `remove` tests

## Purpose

Adds default parameters to `GenericCollection#delete` and introduces `GenericCollection#deleteOne` and `GenericCollection#deleteMany`. I also changed the tests to not use deprecated methods

## Background Context

I tried to replicate the default parameters for `GenericCollection#delete` as described in https://docs.mongodb.com/manual/reference/command/delete/
I introduced `GenericCollection#deleteOne` and `GenericCollection#deleteMany` as described in https://docs.mongodb.com/manual/reference/method/db.collection.deleteOne/ and https://docs.mongodb.com/manual/reference/method/db.collection.deleteMany/ as syntax sugar over the collection API

## References

https://groups.google.com/forum/#!topic/reactivemongo/9Sjq-qQg61Y

### DIFF
--- a/driver/src/main/scala/api/collections/DeleteOps.scala
+++ b/driver/src/main/scala/api/collections/DeleteOps.scala
@@ -55,7 +55,8 @@ trait DeleteOps[P <: SerializationPack with Singleton]
     protected def bulkRecover: Option[Exception => Future[WriteResult]]
 
     /**
-     * Performs a delete with a single selector (see [[BatchCommands.DeleteCommand.DeleteElement]]).
+     * Performs a delete with a one single selector (see [[BatchCommands.DeleteCommand.DeleteElement]]).
+     * This will delete all the documents matched by the `q` selector.
      */
     final def one[Q, U](q: Q, limit: Option[Int] = None, collation: Option[Collation] = None)(implicit ec: ExecutionContext, qw: pack.Writer[Q]): Future[WriteResult] = element[Q, U](q, limit, collation).flatMap { upd => execute(Seq(upd)) }
 
@@ -69,13 +70,13 @@ trait DeleteOps[P <: SerializationPack with Singleton]
       }
 
     /**
-     * Deletes documents, according to the [[DeleteElement]]s.
+     * Performs a bulk operation using many deletes, each can delete multiple documents.
      *
      * {{{
      * import reactivemongo.bson.BSONDocument
      * import reactivemongo.api.collections.BSONCollection
      *
-     * def deleteMany(coll: BSONCollection, docs: Iterable[BSONDocument]) = {
+     * def bulkDelete(coll: BSONCollection, docs: Iterable[BSONDocument]) = {
      *   val delete = coll.delete(ordered = true)
      *   val elements = docs.map { doc =>
      *     delete.element(
--- a/driver/src/main/scala/api/collections/DeleteOps.scala
+++ b/driver/src/main/scala/api/collections/DeleteOps.scala
@@ -55,7 +55,7 @@ trait DeleteOps[P <: SerializationPack with Singleton]
     protected def bulkRecover: Option[Exception => Future[WriteResult]]
 
     /**
-     * Performs a [[https://docs.mongodb.com/manual/reference/method/db.collection.deleteOne/ single delete]] (see [[BatchCommands.DeleteCommand.DeleteElement]]).
+     * Performs a delete with a single selector (see [[BatchCommands.DeleteCommand.DeleteElement]]).
      */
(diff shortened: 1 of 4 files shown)


driver/src/main/scala/api/collections/GenericCollection.scala
```diff
@@ -622,6 +622,14 @@ trait GenericCollection[P <: SerializationPack with Singleton] extends Collectio
       Future.failed(MissingMetadata())
   }
 
+  def deleteOne[S](selector: S, writeConcern: WriteConcern = defaultWriteConcern)(implicit swriter: pack.Writer[S], ec: ExecutionContext): Future[WriteResult] = {
```
cchantep: I prefer not to introduce those aliases for `delete.{one,many}`

driver/src/main/scala/api/gridfs.scala
```diff
@@ -462,9 +462,9 @@ class GridFS[P <: SerializationPack with Singleton](db: DB with DBMetaCommands,
    * @param id the file id to remove from this store
    */
   def remove[Id <: pack.Value](id: Id)(implicit ctx: ExecutionContext, idProducer: IdProducer[Id]): Future[WriteResult] =
-    asBSON(chunks.name).remove(BSONDocument(idProducer("files_id" -> id))).
+    asBSON(chunks.name).deleteMany(BSONDocument(idProducer("files_id" -> id))).
```
cchantep: `.delete.one`

driver/src/test/scala/CollectionSpec.scala
```diff
@@ -235,7 +235,7 @@ class CollectionSpec(implicit protected val ee: ExecutionEnv)
         "name" -> BSONJavaScript("db.getName()"))).flatMap { _ =>
         find.map(_.flatMap(_.getAs[BSONJavaScript]("name")).map(_.value))
       } aka "inserted" must beSome("db.getName()").await(1, timeout) and {
-        collection.remove(selector).map(_.n) must beTypedEqualTo(1).
+        collection.deleteMany(selector).map(_.n) must beTypedEqualTo(1).
```
cchantep: `delete.one`

driver/src/main/scala/api/collections/GenericCollection.scala
jtjeferreira: please note that `def deleteMany` is not an alias for `delete.many`, but an alias for `delete.one` without a limit on the documents to delete. There is no alias for `delete.many`.
Having to call `delete().one(selector, limit = None)` to delete all (i.e many) documents that match the selector seems a bit unintuitive to me... 
If you still think that the best is to not having these methods, I will gladly remove them and had them in my project as extension methods...

jtjeferreira: :ok_hand:

cchantep: So these could be removed for now. Thx

driver/src/main/scala/api/gridfs.scala
```diff
@@ -462,9 +462,9 @@ class GridFS[P <: SerializationPack with Singleton](db: DB with DBMetaCommands,
    * @param id the file id to remove from this store
    */
   def remove[Id <: pack.Value](id: Id)(implicit ctx: ExecutionContext, idProducer: IdProducer[Id]): Future[WriteResult] =
-    asBSON(chunks.name).deleteMany(BSONDocument(idProducer("files_id" -> id))).
+    asBSON(chunks.name).delete().one(BSONDocument(idProducer("files_id" -> id)), limit = None).
```
cchantep: `limit` is defaulted `None` on `def one`

driver/src/main/scala/api/collections/GenericCollection.scala
jtjeferreira: removed

driver/src/main/scala/api/collections/DeleteOps.scala
```diff
@@ -55,7 +55,7 @@ trait DeleteOps[P <: SerializationPack with Singleton]
     protected def bulkRecover: Option[Exception => Future[WriteResult]]
 
     /**
-     * Performs a [[https://docs.mongodb.com/manual/reference/method/db.collection.deleteOne/ single delete]] (see [[BatchCommands.DeleteCommand.DeleteElement]]).
+     * Performs a delete with a single selector (see [[BatchCommands.DeleteCommand.DeleteElement]]).
```
jtjeferreira: removed the reference to deleteOne because this method can be used for both `deleteOne` and `deleteMany`

cchantep:
Thx for the contribution

PR #759, opened 2018-07-11
Correcting Log4J direct usage in compile scope.

alexflav23:
# Pull Request Checklist

## Fixes

Fixes direct usage of Log4J in the compile scope of the project. It's not a good pattern to enforce a specific logging implementation onto consumers of ReactiveMongo, and no implementation details should rely on a specific implementation of the SLF4J facade.

## Purpose

Removes directly dependency on Log4J inside the compile scope, and moves it to test scope only instead. `LazyLogging` correctly delegates to the logging facade now.

## Background Context

Lots of users, ourselves included, use ReactiveMongo widely throughout projects. For various, perhaps legacy concerns, we cannot pick Log4J2 and instead rely on Logback. ReactiveMongo should not be, as a library, the piece of software forcing us down one path or the other, as this causes ReactiveMongo specific logging to completely breakdown and go missing in our projects.

### DIFF
--- a/driver/src/main/scala/utils.scala
+++ b/driver/src/main/scala/utils.scala
@@ -46,7 +46,7 @@ object `package` {
 
 @deprecated(
   message = "Use [[reactivemongo.util.LazyLogger]]", since = "0.12.0")
-case class LazyLogger(logger: org.apache.logging.log4j.Logger) {
+case class LazyLogger(logger: org.slf4j.Logger) {
   def trace(s: => String): Unit = { if (logger.isTraceEnabled) logger.trace(s) }
   def trace(s: => String, e: => Throwable): Unit = { if (logger.isTraceEnabled) logger.trace(s, e) }
   def debug(s: => String): Unit = { if (logger.isDebugEnabled) logger.debug(s) }
@@ -63,7 +63,7 @@ case class LazyLogger(logger: org.apache.logging.log4j.Logger) {
   message = "Use [[reactivemongo.util.LazyLogger]]", since = "0.12.0")
 object LazyLogger {
   def apply(logger: String): LazyLogger =
-    LazyLogger(org.apache.logging.log4j.LogManager.getLogger(logger))
+    LazyLogger(org.slf4j.LoggerFactory.getLogger(logger))
 }
 
 @deprecated(
--- a/project/ReactiveMongo.scala
+++ b/project/ReactiveMongo.scala
@@ -130,20 +130,19 @@ object Dependencies {
     "org.specs2" %% "specs2-core" % specsVer.value % Test
   }
 
-  val slf4jVer = "1.7.12"
+  val slf4jVer = "1.7.21"
   val log4jVer = "2.5"
 
   val slf4j = "org.slf4j" % "slf4j-api" % slf4jVer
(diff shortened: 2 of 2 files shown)


driver/src/main/scala/utils.scala
```diff
@@ -46,7 +46,7 @@ object `package` {
 
 @deprecated(
   message = "Use [[reactivemongo.util.LazyLogger]]", since = "0.12.0")
-case class LazyLogger(logger: org.apache.logging.log4j.Logger) {
```
cchantep: Deprecated so no change

project/ReactiveMongo.scala
```diff
@@ -130,20 +130,19 @@ object Dependencies {
...
 
-  val slf4jVer = "1.7.12"
+  val slf4jVer = "1.7.21"
   val log4jVer = "2.5"
 
   val slf4j = "org.slf4j" % "slf4j-api" % slf4jVer
-  val slf4jSimple = "org.slf4j" % "slf4j-simple" % slf4jVer
 
   val logApi = Seq(
-    slf4j % "provided",
```
cchantep: Not desired

driver/src/main/scala/utils.scala
alexflav23: @cchantep That doesn't make much of a difference, I'm happy to remove this if it's deprecated instead. I noticed there's already an SLF4J one. Unfortunately, this setup is causing issues now.

project/ReactiveMongo.scala
alexflav23: @cchantep Happy to return to a provided dependency, however you should consider removing the Log4J one as soon as, because it's causing logs to fail for most clients, only partial logs show.

driver/src/main/scala/utils.scala
cchantep: By nature, a deprecated member should not be changed, but in the end removed

project/ReactiveMongo.scala
cchantep: - Most of these dependencies are `Test`-only, so there is no impact on the runtime of the project using the lib.
- The `provided` semantic is appropriate for `slf4j` (and is on-purpose)

.gitignore
```diff
@@ -12,3 +12,4 @@ target/
 .ci_scripts/signing-ca.csr
 .ci_scripts/signing-ca.key
 .ci_scripts/signing-ca
+.idea/
```
cchantep: IDE specific => `.git/info/exclude`

project/Dependencies.scala
```diff
@@ -41,17 +41,18 @@ object Dependencies {
...
+
   val slf4j = "org.slf4j" % "slf4j-api" % slf4jVer
-  val slf4jSimple = "org.slf4j" % "slf4j-simple" % slf4jVer
 
   val logApi = Seq(
     slf4j % "provided",
-    "org.apache.logging.log4j" % "log4j-api" % log4jVer // deprecated
-  ) ++ Seq("log4j-core", "log4j-slf4j-impl").map(
-    "org.apache.logging.log4j" % _ % log4jVer % Test)
+    "org.apache.logging.log4j" % "log4j-api" % log4jVer % Test,
```
cchantep: Not needed, would rather add `log4j-api` to the `Seq` form

PR #762, opened 2018-07-15
Awaiting channel

cchantep:


driver/src/main/scala/api/cursor.scala
```diff
@@ -687,11 +615,18 @@ object DefaultCursor {
...
         logger.debug(s"[$logCat] Clean up $cursorID, sending KillCursors")
 
-        val killReq = RequestMaker(
-          KillCursors(Set(cursorID)),
-          readPreference = preference)
+        val killReq = RequestMakerExpectingResponse(
+          RequestMaker(KillCursors(Set(cursorID)), readPreference = preference),
+          false)
+
+        import ExecutionContext.Implicits.global
```
jtjeferreira: ExecutionContext.Implicits.global

cchantep: That's WIP

jtjeferreira: :+1:

eugeneatnezasa:
Strange behavior if I set nbChannelsPerNode=1, occupy this channel by slow query and then send several fast queries:
several fast queries fail (No active channel ...), several executed successfully
And everything happens after slow query completed

cchantep:
Which is coherent, if no channel is available in the pool, it fail. Alternative would be to queue request.

cchantep:
@eugeneatnezasa unless you've remove the failover, this is used to retry (and possibly succeed).

eugeneatnezasa:
sorry, it was a mistake with approval (mixed up multiple github windows)

eugeneatnezasa:
@cchantep I approved twice this PR by mistake (wrong window). Just notifying

cchantep:
@eugeneatnezasa You're free to approve if it's ok for you ;-)

PR #765, opened 2018-07-27
Session management

cchantep:
https://github.com/mongodb/specifications/blob/master/source/causal-consistency/causal-consistency.rst#causally-consistent-read-commands


driver/src/main/scala/api/collections/GenericCollection.scala
```diff
@@ -131,6 +138,7 @@ trait GenericCollection[P <: SerializationPack with Singleton] extends Collectio
    * @param swriter $swriterParam
    * @return $returnQueryBuilder
    */
+  @deprecated("Use `find` with optional `projection`", "0.16.0")
```
jtjeferreira: can this deprecation be reverted? or maybe remove this method so that usages of find with only 1 parameter will use `def find[S, J](selector: S, projection: Option[J] = Option.empty)` below?


I was trying the new 0.16 release and have tons of warning because of this change. And I don't see an easy way to migrate...

cchantep: This is required to prepare API changes, warning what changes will come. So this won't be reverted.

You can ignore use deprecation warning for now, or add `, Option.empty[BSONDocument]`.

cchantep: Deprecation is something we have to deal with, unfortunately Scalac is not very good at filtering during API migration (e.g. https://github.com/scala/bug/issues/1781 ).

Anyway this won't be reverted and the projection parameter of the new function cannot be defaulted, or it will be conflicting.

If you cannot either add the explicit parameter or suppress warning (see [silencer](https://github.com/ghik/silencer/blob/master/README.md)), then it only let you the choice not to upgrade until a future version where API would be cleanup (not precise roadmap, but probably not the next 0.17).

Best regards

jtjeferreira: Thanks for that silencer link, I was only aware of this [workaround](https://issues.scala-lang.org/browse/SI-7934?focusedCommentId=66403&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-66403), but again I would have to add it to the 500 call sites :thinking:

jtjeferreira: > the projection parameter of the new function cannot be defaulted, or it will be conflicting.

What do you mean? the `projection` parameter of the new `find` function has a default of `Option.empty. Or you meant another function?

jtjeferreira: hi @cchantep! just to let you know that I found a workaround, by copying this class to my project and removing the deprecated annotations from the methods I was using. Then I used the silencer plugin to suppress warnings from this class. This was more practical than forking the whole project and publishing somewhere...


Another issue I had, was with `reactivemongo.api.commands#WriteConcern`. Since this was defined in a package object I could not do what I described above. However I managed to remove all the usages of WriteConcern, because I was always using `Acknowledged` which is the default...

I really want to try this version to test the fix in #762. Will get to back to you later with the results...

jbgi: Btw, I am also hit by this change, and adding `, projection = None` does not solve every case because it can cause ambiguous implicit resolution. cf. https://github.com/ReactiveMongo/ReactiveMongo/pull/770

ps: the new api is not used at all in tests, would probably be good to migrate test to new APIs to avoid this kind of usability issues.

cchantep: `None` is not good for inference, so is it for implicit resolution, use `Option.empty[BSONDocument]`

cchantep: @jtjeferreira the `WriteConcern` has changed in 0.16.0 (but it's planned to "normalized" it in next releases).

See [silencer example](https://github.com/cchantep/ReactiveMongo/tree/0.16.0/silencer-example) (note that it's using https://github.com/cchantep/tatami as resolver)

jbgi: > None is not good for inference, so is it for implicit resolution, use Option.empty[BSONDocument]

Yeah I come to this conclusion after writing this. Still it is quite verbose, and not very intuitive.

Another case where the new API is  verbose is `distinct`, for which I had to add three new parameters: `, query = None, ReadConcern.Local, collation = None`.
Or maybe those parameters are missing defaults values?

cchantep: That's the same, default values cannot be added without conflicting, and would only be added once the API cleanup is ok for it.

jtjeferreira: > the WriteConcern has not changed in 0.16.0 (but it's planned to "normalized" it in next releases).

I meant to say that `WriteConcern` was also deprecated in this PR, and there is no way to use it without deprecation warnings (except by using the silencer). I solved it in my code base by not using it :)

PR #770, opened 2018-08-07
ambiguous implicits with new find api (0.16).

jbgi:
This fails at compile with
```
[error] /home/giraudeau/external-scm/ReactiveMongo/driver/src/test/scala/CollectionSpec.scala:47:24: ambiguous implicit values:
[error]  both object BSONDocumentIdentity in trait DefaultBSONHandlers of type reactivemongo.bson.package.BSONDocumentIdentity.type
[error]  and value writer of type CollectionSpec.this.PersonWriter.type
[error]  match expected type CollectionSpec.this.slowColl.pack.Writer[J]
[error]           slowColl.find(BSONDocument.empty, projection = None).cursor[BSONDocument]().
[error]                        ^
```


jbgi:
fixed by using `Option.empty[BSONDocument]` instead of `None`.

PR #771, opened 2018-08-11
BSON Handlers improvements

jtjeferreira:
# Pull Request Checklist

* [X] Have you added tests for any changed functionality? No changed functionality and current test still pass

## Purpose

improves compilation speed by making implicit search for BSONReaders and BSONWriters faster

## Background Context

This is the same approach used in other libraries (play-json, circe)

### DIFF
--- a/bson/src/main/scala/DefaultBSONHandlers.scala
+++ b/bson/src/main/scala/DefaultBSONHandlers.scala
@@ -30,7 +30,7 @@ class VariantBSONReaderWrapper[B <: BSONValue, T](reader: VariantBSONReader[B, T
   def read(b: B) = reader.read(b)
 }
 
-trait DefaultBSONHandlers {
+trait DefaultBSONHandlers extends LowPrioBSONHandlers {
   import scala.language.higherKinds
 
   implicit object BSONIntegerHandler extends BSONHandler[BSONInteger, Int] {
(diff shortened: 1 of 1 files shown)


bson/src/main/scala/DefaultBSONHandlers.scala
```diff
@@ -120,24 +120,6 @@ trait DefaultBSONHandlers {
   implicit def bsonBooleanLikeReader[B <: BSONValue] =
     new BSONBooleanLikeReader[B]
 
-  // Collections Handlers
```
cchantep: I don't see the goal, those instances are not low priority for me

jtjeferreira: I can prepare some examples with "-Xlog-implicits" so you can see the compiler trying to use BSONArrayCollectionWriter where it is not needed, but like I said above other libraries to it like this: [play-json](https://github.com/playframework/play-json/blob/master/play-json/shared/src/main/scala/play/api/libs/json/Reads.scala#L186) and [circe](https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Decoder.scala#L281)

cchantep: There are other reason why Play JSON is doing it in this way (I'm contrib on both).

jtjeferreira:
I will prepare the examples then. Maybe later today...

bson/src/main/scala/DefaultBSONHandlers.scala
jtjeferreira: I will prepare the examples then. Maybe later today

cchantep: thx

bson/src/main/scala/DefaultBSONHandlers.scala
```diff
@@ -184,9 +166,6 @@ trait DefaultBSONHandlers {
   implicit def findWriter[T](implicit writer: VariantBSONWriter[T, _ <: BSONValue]): BSONWriter[T, _ <: BSONValue] =
     new VariantBSONWriterWrapper(writer)
 
-  implicit def findReader[T](implicit reader: VariantBSONReader[_ <: BSONValue, T]): BSONReader[_ <: BSONValue, T] =
```
jtjeferreira: please note that I also remove this implicit, because I think it is never used...

PR #784, opened 2018-11-15
Watch collection change stream

eleaar:
This pull request mainly introduces support for watching the changeStream of a collection on Mongo 3.6+.

Other things:
- I also added support for the `maxTimeMS` parameter on aggregate commands.
- I added some utilities in tests to work with a temporary db/collection in isolation

### DIFF
--- a/driver/src/main/scala/api/ChangeStreams.scala
+++ b/driver/src/main/scala/api/ChangeStreams.scala
@@ -1,10 +1,23 @@
 package reactivemongo.api
 
 object ChangeStreams {
+
+  /**
+   * Defines the lookup strategy of a change stream.
+   */
   sealed abstract class FullDocument(val name: String)
 
   object FullDocument {
+
+    /**
+     * Default lookup strategy. Insert and Replace events contain the full document at the time of the event.
+     */
     case object Default extends FullDocument("default")
+
+    /**
+     * In this strategy, in addition to the default behavior, Update change events will be joined with the *current*
+     * version of the related document (which is thus not necessarily the value at the time of the event).
+     */
     case object UpdateLookup extends FullDocument("updateLookup")
   }
 }
--- a/driver/src/main/scala/api/ChangeStreams.scala
+++ b/driver/src/main/scala/api/ChangeStreams.scala
@@ -1,4 +1,4 @@
-package api
+package reactivemongo.api
 
 object ChangeStreams {
   sealed abstract class FullDocument(val name: String)
@@ -7,5 +7,4 @@ object ChangeStreams {
     case object Default extends FullDocument("default")
     case object UpdateLookup extends FullDocument("updateLookup")
   }
-
 }
--- a/driver/src/main/scala/api/ChangeStreams.scala
+++ b/driver/src/main/scala/api/ChangeStreams.scala
@@ -1,0 +1,11 @@
+package api
+
+object ChangeStreams {
+  sealed abstract class FullDocument(val name: String)
+
+  object FullDocument {
+    case object Default extends FullDocument("default")
+    case object UpdateLookup extends FullDocument("updateLookup")
+  }
+
+}
--- a/driver/src/main/scala/api/ChangeStreams.scala
+++ b/driver/src/main/scala/api/ChangeStreams.scala
@@ -5,19 +5,19 @@ object ChangeStreams {
   /**
    * Defines the lookup strategy of a change stream.
    */
-  sealed abstract class FullDocument(val name: String)
+  sealed abstract class FullDocumentStrategy(val name: String)
 
-  object FullDocument {
+  object FullDocumentStrategy {
 
     /**
      * Default lookup strategy. Insert and Replace events contain the full document at the time of the event.
      */
-    case object Default extends FullDocument("default")
+    case object Default extends FullDocumentStrategy("default")
 
     /**
      * In this strategy, in addition to the default behavior, Update change events will be joined with the *current*
      * version of the related document (which is thus not necessarily the value at the time of the event).
      */
-    case object UpdateLookup extends FullDocument("updateLookup")
+    case object UpdateLookup extends FullDocumentStrategy("updateLookup")
   }
 }
--- a/driver/src/main/scala/api/collections/ChangeStreamOps.scala
+++ b/driver/src/main/scala/api/collections/ChangeStreamOps.scala
@@ -25,7 +25,7 @@ trait ChangeStreamOps[P <: SerializationPack with Singleton] { collection: Gener
    * @param maxAwaitTimeMS       The maximum amount of time in milliseconds the server waits for new data changes
    *                             before returning an empty batch. In practice, this parameter controls the duration
    *                             of the long-polling behavior of the cursor.
-   * @param fullDocument         if set to UpdateLookup, every update change event will be joined with the *current*
+   * @param fullDocumentStrategy         if set to UpdateLookup, every update change event will be joined with the *current*
    *                             version of the relevant document.
    * @param reader               a reader of the resulting Change Events
    * @tparam T the type into which Change Events are deserialized
@@ -35,11 +35,11 @@ trait ChangeStreamOps[P <: SerializationPack with Singleton] { collection: Gener
     startAtOperationTime: Option[Value] = None,
     pipeline: List[PipelineOperator] = Nil,
     maxAwaitTimeMS: Option[Long] = None,
-    fullDocument: Option[ChangeStreams.FullDocument] = None
+    fullDocumentStrategy: Option[ChangeStreams.FullDocumentStrategy] = None
   )(implicit reader: pack.Reader[T]): WatchBuilder[T] = {
     new WatchBuilder[T] {
       protected val context: AggregatorContext[T] = aggregatorContext[T](
-        firstOperator = new ChangeStream(resumeAfter, startAtOperationTime, fullDocument),
+        firstOperator = new ChangeStream(resumeAfter, startAtOperationTime, fullDocumentStrategy),
         otherOperators = pipeline,
         readConcern = Some(ReadConcern.Majority),
         cursorOptions = CursorOptions.empty.tailable,
@@ -74,6 +74,14 @@ trait ChangeStreamOps[P <: SerializationPack with Singleton] { collection: Gener
      * Unless you are in a fully reactive scenario, you may want to add some timeout behavior to the resulting Future.
      * In that case, remember to explicitly close the cursor when the timeout triggers, so that you don't leak the
      * cursor (the cursor provider needs to support such a functionality).
+     *
+     * 3. New change streams return no data when the cursor is initially established (only subsequent GetMore commands
+     * will actually return the subsequent events). Therefore, such a cursor `head` will always be empty. Only folding
+     * the cursor (directly or through a higher-level cursor provider) will provide the next change event.
+     *
+     * 4. Resumed change streams (via id or operation time) will return the next event when the cursor is initially
+     * established, if there is is some next event. Therefore, `head` is guaranteed to eventually return the next
+     * change event beyond the resume point, when such an event appears.
      */
     def cursor[AC[_] <: Cursor.WithOps[_]](implicit cp: CursorProducer.Aux[T, AC]): AC[T] = {
       context.prepared[AC].cursor
--- a/driver/src/main/scala/api/collections/ChangeStreamOps.scala
+++ b/driver/src/main/scala/api/collections/ChangeStreamOps.scala
@@ -1,0 +1,85 @@
+package reactivemongo.api.collections
+
+import scala.language.higherKinds
+
+import api.ChangeStreams
+import reactivemongo.api._
+
+trait ChangeStreamOps[P <: SerializationPack with Singleton] { collection: GenericCollection[P] =>
+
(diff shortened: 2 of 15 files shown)


driver/src/test/scala/util/WithTemporaryCollection.scala
```diff
@@ -0,0 +1,18 @@
+package util
+
+import scala.util.Random
+
+import reactivemongo.api.{ Collection, CollectionMetaCommands, CollectionProducer }
+import tests.Common.timeout
+
+trait WithTemporaryCollection extends WithTemporaryDb { this: org.specs2.mutable.Specification =>
```
cchantep: What I'm afraid of, is that it tend to make some features not tested in the different kind of env/conf, whereas, even if it should be important in some case in it's (e.g. different behaviour of command according SSL/auth/version, ...).

eleaar: I'm not sure to see how that would do that? You can still run the same tests in multiple environments, right? 
As I see it, this utility just provides an isolated collection when one is needed and acceptable, either at the suite or specification level.

driver/src/main/scala/api/collections/ChangeStreamOps.scala
```diff
@@ -0,0 +1,85 @@
...
+
+import scala.language.higherKinds
+
+import api.ChangeStreams
+import reactivemongo.api._
+
+trait ChangeStreamOps[P <: SerializationPack with Singleton] { collection: GenericCollection[P] =>
+
+  import BatchCommands.AggregationFramework.ChangeStream
+  // I wasn't able to prove that it is equal to `collection.pack.Value`
```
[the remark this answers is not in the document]
eleaar: I admit I lost the battle with the dependent types. :(

Within PipelineOperators, we only have access to types path-dependent on `AggregationPipeline.pack` (such as `AggregationPipeline.pack.Value` into which `PipelineOperator`s are serialized).

Normally, `collection.pack.type =:= collection.BatchCommands.AggregationFramework .pack.type`, but I was unable to prove it. So I used the latter here to type some arguments to the watch method.

From the user's perspective it will have no practical difference, as for any concrete instance of GenericCollection the two above types will be the same.

I can remove the comment, though... :)

driver/src/test/scala/util/WithTemporaryCollection.scala
cchantep: Isolated collection is not an issue, but isolated DB (that for example differ in auth) can be. There is already some place where different DB is used, but we need to be careful not to make it the general rule.

eleaar: Ah, I see your point. I will parameterize it on the db, how about that?

cchantep: `Common.databases` can be used, it also include resolution of database through the slow proxy (if needed I would prefer plain function, without inheritance).

eleaar: I have parameterized the temporary collection so that each tests can choose whether to run against the common db, the slow one, or both.

I could make temporaryCollections a plain function. But for the db - it needs to be teared down after the tests, so the loaner pattern will not be enough and I think a mixin is more practical.
But then maybe it's better to have a common style for both?

driver/src/test/scala/util/WithTemporaryCollection.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  protected implicit def ee: ExecutionEnv
+
+  protected final def withTmpCollection[C <: Collection with CollectionMetaCommands, A](
+    db: DefaultDB)(
+    f: C => A)(implicit producer: CollectionProducer[C]): A = {
+    val collectionName = s"tmp-${System identityHashCode this}-${Random.alphanumeric.take(10).mkString("")}"
+    val collection = db[C](collectionName)
+    // we won't drop the collection in case of exceptions, so that it can be debugged
+    Await.ready(collection.create(), timeout)
```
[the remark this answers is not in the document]
eleaar: That's what I would do in scalatest, but as I see it, specs2 is more oriented toward synchronous testing (you lift your matchers, not your match results). And I don't have an identity monad to just parameterize in a tagless-way

driver/src/test/scala/ChangeStreamSpec.scala
```diff
@@ -0,0 +1,205 @@
...
+          "foo" -> "baz",
+        )
+
+        // when
+        val firstEventFuture = foldOne(cursor)
+        // See comment above
+        delayBy(500.millis) {
+          coll.insert(ordered = false).many(Seq(testDocument1, testDocument2))
+        } must beEqualTo(()).await
+        val result = firstEventFuture.flatMap { firstEvent =>
```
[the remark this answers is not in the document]
eleaar: I don't understand?

driver/src/test/scala/util/WithTemporaryCollection.scala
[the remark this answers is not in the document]
cchantep: Specs2 is handling async with `.await` as scalatest. Most of the integration tests there are using it.

driver/src/test/scala/ChangeStreamSpec.scala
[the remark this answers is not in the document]
cchantep:
```scala
x must y and {
  z must w
}
```

driver/src/main/scala/api/ChangeStreams.scala
```diff
@@ -0,0 +1,23 @@
+package reactivemongo.api
+
+object ChangeStreams {
+
+  /**
+   * Defines the lookup strategy of a change stream.
+   */
+  sealed abstract class FullDocument(val name: String)
```
cchantep: Maybe `Strategy` as `FailoverStrategy` ?

driver/src/main/scala/api/commands/AggregationFramework.scala
```diff
@@ -1,6 +1,6 @@
 package reactivemongo.api.commands
 
-import reactivemongo.api.{ ReadConcern, SerializationPack }
+import reactivemongo.api.{ChangeStreams, ReadConcern, SerializationPack}
```
cchantep: Formatting issue

driver/src/test/scala/ChangeStreamSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+  private val actorSystem = ActorSystem("changeStreams")
+
+  override def afterAll: Unit = {
+    Await.ready(actorSystem.terminate(), timeout)
+    super.afterAll
+  }
+
+  private def foldOne[T](cursor: Cursor.WithOps[T]): Future[T] = {
+    cursor.collect[List](maxDocs = 1, Cursor.FailOnError()).flatMap { result =>
+      Future.fromTry(Try(result.head))
```
cchantep: Will break the 2.10 build (not yet dropped) I guess

driver/src/test/scala/ChangeStreamSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+        }.await(retries = 2, 1.second)
+      }
+    }
+  }
+
+  private def skipIfNotRSAndNotVersionAtLeast[R: AsResult](version: MongoWireVersion)(r: => R) = {
+    skippedIf(isNotReplicaSet, isNotAtLeast(db, version))(r)
+  }
+
+  private val actorSystem = ActorSystem("changeStreams")
```
cchantep: A separate one is required? Otherwise a system is available in `Common`

driver/src/test/scala/ChangeStreamSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+
+        // then
+        result must beLike[BSONDocument] {
+          case event =>
+            event.getAs[String]("operationType") must beSome("update")
+            event.getAs[BSONDocument]("documentKey")
+              .flatMap(_.getAs[String]("_id")) must beSome(id)
+            event.getAs[BSONDocument]("fullDocument")
+              .flatMap(_.getAs[String](fieldName)) must beSome(lastValue)
+        }.await(retries = 2, 1.second)
```
cchantep: Rather use the default timeout (updated accordingly on the CI) ?

driver/src/main/scala/api/commands/AggregationFramework.scala
```diff
@@ -522,4 +522,40 @@ trait AggregationFramework[P <: SerializationPack]
...
+   *
+   * @param resumeAfter the id of the last known Change Event, if any. The stream will resume just after that event.
+   * @param startAtOperationTime the operation time before which all Change Events are known. Must be in the time range
+   *                             of the oplog. (since MongoDB 4.0)
+   * @param fullDocument if set to UpdateLookup, every update change event will be joined with the *current* version of the
+   *                     relevant document.
+   */
+  final class ChangeStream(
+    resumeAfter: Option[pack.Value] = None,
+    startAtOperationTime: Option[pack.Value] = None, // TODO restrict to something more like a timestamp?
```
cchantep: What's expected by the command? A bson timestamp? If yes we can require a `Long` ? Or a `T: Numeric` ?

driver/src/test/scala/ChangeStreamSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  private val actorSystem = ActorSystem("changeStreams")
+
+  override def afterAll: Unit = {
+    Await.ready(actorSystem.terminate(), timeout)
+    super.afterAll
+  }
+
+  private def foldOne[T](cursor: Cursor.WithOps[T]): Future[T] = {
```
cchantep: `cursor.head` ?

driver/src/test/scala/ChangeStreamSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+        val result = foldOne(cursor)
+        // The cursor needs to have been opened by the time the insert operation is executed
+        // We have no means to guarantee that with the Cursor API, so the following is a best-effort solution,
+        // but a race-condition is still possible and may lead to a false negative test result.
+        delayBy(500.millis) {
+          coll.insert(ordered = false).one(testDocument)
+        } must beEqualTo(()).await
+
+        // then
+        result must beLike[BSONDocument] {
```
cchantep: `and`

driver/src/test/scala/ChangeStreamSpec.scala
eleaar: precisely not :) `headOption` will alway return `None` on a change stream cursor

driver/src/test/scala/ChangeStreamSpec.scala
cchantep: `Common.driver.system` (as in ChannelFactorySpec`)

driver/src/main/scala/api/commands/AggregationFramework.scala
cchantep: If it can be opaque with responsibility to the dev/caller to provide something appropriate (doc). In this case would rename the arg without time which tend to "force" the expected semantic/type, maybe `offset` or a name like that.

driver/src/test/scala/util/WithTemporaryCollection.scala
```diff
@@ -0,0 +1,21 @@
+package util
+
+import scala.concurrent.{ Await, ExecutionContext }
```
cchantep:
```suggestion
import scala.concurrent.{ Await, ExecutionContext }
import scala.concurrent.duration.FiniteDuration

```

driver/src/test/scala/util/WithTemporaryCollection.scala
```diff
@@ -0,0 +1,21 @@
...
+import scala.util.Random
+
+import reactivemongo.api.{ Collection, CollectionMetaCommands, CollectionProducer, DefaultDB }
+import tests.Common.timeout
+
+object WithTemporaryCollection {
+
+  def withTmpCollection[C <: Collection with CollectionMetaCommands, A](db: DefaultDB)(f: C => A)(implicit producer: CollectionProducer[C], ec: ExecutionContext): A = {
```
cchantep:
```suggestion
  def withTmpCollection[C <: Collection with CollectionMetaCommands, A](
    db: DefaultDB,
    timeout: FiniteDuration = tests.Common.timeout)(f: C => A)(implicit producer: CollectionProducer[C], ec: ExecutionContext): A = {
```

driver/src/main/scala/api/commands/AggregationFramework.scala
eleaar: I tried to keep the name the same as in the official API, so its semantics are implied by the general mongo doc.

driver/src/test/scala/ChangeStreamSpec.scala
eleaar: `cursor.head` will send a single `Aggregate` command which will always return no elements in this case, unless the stream is being resumed.

`cursor.collect(maxDocs = 1)` will send the initial `Aggregate` command, and then will continue sending `GetMore` commands until a new event is available.

driver/src/test/scala/util/WithTemporaryCollection.scala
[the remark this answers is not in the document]
eleaar: to use `await` I would need to make it a mixin again, wouldn't I?

PR #785, opened 2018-12-05
Reconnect after a configurable delay

konradwudkowski:
# Pull Request Checklist

* [ x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [ x] Have you read through the [contributor guidelines](https://github.com/ReactiveMongo/ReactiveMongo/blob/master/CONTRIBUTING.md#reactivemongo-developer--contributor-guidelines)?
* [ x] Have you added tests for any changed functionality?

## Fixes

This fixes performance problems when one of the nodes in the replica set is down and RM tries to reconnect.

## Purpose

This PR introduces a non-blocking, configurable delay for reconnecting when a node is down.

## Background Context

The current version of RM since 0.15.1+ has a performance problem with reconnecting. It tries to reconnect as soon as connection fails. Then reconnect fails again and it goes in a loop as long as a node is down. Reconnecting attempts can happen 1000s times per second impacting the performance of the entire app and maxing out cpu. This is regardless of logging level that is set. A better approach in my opinion is to retry after some delay. I have arbitrarily chosen 1000 ms but I'm not married to this number and the PR lets users choose incl 0ms if they want current behaviour.

This problem seems to be system specific. I was able to reproduce on several linux boxes but problem doesn't happen on Macs, at least on Macs I tested. I don't know what causes the problem or difference between systems but after the fix everything continues to work fine on Macs and also works on linux.

ReactiveMongo generally doesn't do anything platform specific but it uses Netty which has a history of problems with specific linux kernel versions, e.g. https://github.com/netty/netty/issues/2616.

## References

Are there any relevant issues / PRs / mailing lists discussions?
https://groups.google.com/forum/#!topic/reactivemongo/UK26oAL4eIo

### DIFF
--- a/driver/src/main/scala/core/nodeset/ChannelFactory.scala
+++ b/driver/src/main/scala/core/nodeset/ChannelFactory.scala
@@ -1,5 +1,6 @@
 package reactivemongo.core.nodeset // TODO: Move to `netty` package
 
+import java.util.concurrent.TimeUnit
 import java.lang.{ Boolean => JBool }
 
 import scala.concurrent.Promise
@@ -71,7 +72,12 @@ private[reactivemongo] final class ChannelFactory(
               s"Connection to ${host}:${port} refused for channel #${chanId}",
               op.cause)
 
-            receiver ! ChannelDisconnected(chanId)
+            op.channel().eventLoop().schedule(new Runnable {
+              def run(): Unit = {
+                receiver ! ChannelDisconnected(chanId)
+              }
+            }, options.reconnectDelayMS, TimeUnit.MILLISECONDS)
+
           }
         }
       })
--- a/driver/src/main/scala/api/MongoConnection.scala
+++ b/driver/src/main/scala/api/MongoConnection.scala
@@ -633,6 +633,9 @@ object MongoConnection {
           case ("rm.nbChannelsPerNode", v) => unsupported -> result.
             copy(nbChannelsPerNode = v.toInt)
 
+          case ("rm.reconnectDelayMS", v) => unsupported -> result.
+            copy(reconnectDelayMS = v.toInt)
+
           case ("writeConcern", "unacknowledged") => unsupported -> result.
             copy(writeConcern = WC.Unacknowledged)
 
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -12,3 +12,4 @@ target/
 .ci_scripts/signing-ca.csr
 .ci_scripts/signing-ca.key
 .ci_scripts/signing-ca
+.idea
```
cchantep: IDE specific should not be committed

driver/src/main/scala/core/nodeset/ChannelFactory.scala
```diff
@@ -71,7 +72,12 @@ private[reactivemongo] final class ChannelFactory(
               s"Connection to ${host}:${port} refused for channel #${chanId}",
               op.cause)
 
-            receiver ! ChannelDisconnected(chanId)
+            op.channel().eventLoop().schedule(new Runnable {
```
cchantep: I don't see the purpose, moreover scheduled/async task should rather go through Akka.

konradwudkowski: The purpose of this is to **not immediately** retry. Current implementation retries immediately when a connection is refused, fails very quickly again, retries again and does this fail-retry cycle **hundreds or thousands times per second** for every channel. There is no reason to retry like this. This impacts the performance of the app since `0.15.1`. CPU is maxed out.  I have provided steps to reproduce it in the referenced mailing list.

`ChannelFactory` currently has no access to the `ActorSystem` or Akka `Scheduler` to schedule a task. If you want I can make a series of changes to constructors of `ChannelFactory`, `StandardDBSystem`, `LegacyDBSystem` etc and pass the `ActorSystem` from within `reactivemongo.api.Driver` or if we don't want to pass the `ActorSystem` I can pass an instance of e.g. `trait Scheduler { def scheduleOnce(delay: FiniteDuration)(block => Unit): Unit }` that I can instantiate using the `ActorSystem` from `reactivemongo.api.Driver`. 

Taking advantage of netty's event loop and scheduler seems simpler in this case and achieves the same purpose but I don't mind scheduling with Akka if you insist at the cost of more changes. Please let me know.

.gitignore
konradwudkowski: removed

driver/src/main/scala/core/nodeset/ChannelFactory.scala
konradwudkowski: hi @cchantep thanks for taking a look and improving this. I've tried your PR #787 and it works well for me so I will close my PR in favour of yours now.

konradwudkowski:
Closing as #787 fixes the problem in a better way

PR #802, opened 2019-02-23
Accept null as none

AurelienRichez:
# Pull Request Checklist

## Fixes

Fixes Option parsing when the key is null.

## Purpose

This PR allows to be more lenient when parsing a null which we expect to be an option. `null` is now accepted as `None` value.

## Background Context

in `0.16.0`, any error encountered while parsing an option resulted in a None. This was fixed to fail when the key actually contains something which does not match the expected type. 

However, having a null also fails now, which is probably not what we expect. 

## References

 - Original topic on the mailing list : https://groups.google.com/forum/?nomobile=true#!topic/reactivemongo/o7coQZEniqw
 - PR which solved the "every error interpreted as None" behaviour : #781

### DIFF
--- a/bson/src/main/scala/types.scala
+++ b/bson/src/main/scala/types.scala
@@ -996,9 +996,11 @@ case class BSONDocument(stream: Stream[Try[BSONElement]])
    */
   def getAsTry[T](key: String)(implicit reader: BSONReader[_ <: BSONValue, T]): Try[T] = {
     val tt = getTry(key)
-    tt.flatMap { element: BSONValue =>
-      Try(reader.asInstanceOf[BSONReader[BSONValue, T]].read(element))
-        .recoverWith { case _ if element == BSONNull => Failure(DocumentKeyNotFound(key)) }
+    tt.flatMap {
+      case BSONNull =>
+        Failure(DocumentKeyNotFound(key))
+      case element =>
+        Try(reader.asInstanceOf[BSONReader[BSONValue, T]].read(element))
     }
   }
 
--- a/bson/src/main/scala/types.scala
+++ b/bson/src/main/scala/types.scala
@@ -17,7 +17,7 @@ package reactivemongo.bson
 
 import java.math.{ BigDecimal => JBigDec }
 
-import exceptions.{ DocumentKeyNotFound, DocumentKeyIsNull }
+import exceptions.DocumentKeyNotFound
 import scala.util.{ Failure, Success, Try }
 import buffer._
 import utils.Converters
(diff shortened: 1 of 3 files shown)


bson/src/main/scala/types.scala
```diff
@@ -996,7 +996,10 @@ case class BSONDocument(stream: Stream[Try[BSONElement]])
    */
   def getAsTry[T](key: String)(implicit reader: BSONReader[_ <: BSONValue, T]): Try[T] = {
     val tt = getTry(key)
-    tt.flatMap { element => Try(reader.asInstanceOf[BSONReader[BSONValue, T]].read(element)) }
+    tt.flatMap { element: BSONValue =>
```
cchantep:
```suggestion
    tt.flatMap {
      case BSONNull =>
        Failure(DocumentKeyIsNull(key))

      case element => 
        Try(reader.asInstanceOf[BSONReader[BSONValue, T]].read(element))
    }
```

bson/src/main/scala/types.scala
```diff
@@ -1006,9 +1009,9 @@ case class BSONDocument(stream: Stream[Try[BSONElement]])
    * If the value could not be deserialized or converted, returns a `Failure`.
    */
   def getAsUnflattenedTry[T](key: String)(implicit reader: BSONReader[_ <: BSONValue, T]): Try[Option[T]] = getAsTry(key)(reader) match {
-    case Failure(_: DocumentKeyNotFound) => Success(None)
-    case Failure(e)                      => Failure(e)
-    case Success(e)                      => Success(Some(e))
+    case Failure(_: DocumentKeyNotFound) | Failure(_: DocumentKeyIsNull) => Success(None)
```
cchantep: Is it really important to have a distinct exception type? Why not `DocumentKeyNotFound`? Would have this change.

bson/src/main/scala/types.scala
AurelienRichez: I used recover to allow the (unlikely) case of a `reader` which actually does something with `null`. This may be far fetched (I never created such a reader myself) so returning direcly an error is probably ok.

bson/src/main/scala/types.scala
AurelienRichez: No, it is not essential. It did this to have a distinction between the 2 cases, but the end result is the same. I will change that

bson/src/main/scala/types.scala
cchantep: People wanting to deal with raw value could do it without reader. I would keep it the simpler.

FYI the [BSON refactoring](https://github.com/cchantep/RM-bson/blob/master/api/src/main/scala/types.scala) avoid such dirty situation.

PR #795, opened 2019-02-15
Fix `IllegalReferenceCountException` in `ResponseDecoder`

avdv:
# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [x] Have you read through the [contributor guidelines](https://github.com/ReactiveMongo/ReactiveMongo/blob/master/CONTRIBUTING.md#reactivemongo-developer--contributor-guidelines)?
* [x] Have you added tests for any changed functionality?

## Purpose

This fixes `io.netty.util.IllegalReferenceCountException`s to be thrown constantly in the `ResponseDecoder`. I noticed these exceptions when debugging [another problem](https://stackoverflow.com/questions/54706942/mongoerror-no-primary-node-is-available-decoderexception) we're seeing regarding "No primary node available" errors.

## Background Context

```
reactivemongo.io.netty.handler.codec.DecoderException: reactivemongo.io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
 	at reactivemongo.io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:98)
 	at reactivemongo.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
 	at reactivemongo.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
 	at reactivemongo.io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
 	at reactivemongo.io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:323)
 	at reactivemongo.io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:297)
 	at reactivemongo.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
 	at reactivemongo.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
 	at reactivemongo.io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
 	at reactivemongo.io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1432)
[...]

PR #803, opened 2019-02-25
Print exceptions in MongoHandler

avdv:
This is a reproducer for PR #795.

### DIFF
--- a/driver/src/main/scala/core/protocol/MongoHandler.scala
+++ b/driver/src/main/scala/core/protocol/MongoHandler.scala
@@ -94,6 +94,7 @@ private[reactivemongo] class MongoHandler(
   }
 
   override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable) = {
+    println(s"ERROR: $cause")
     log(ctx, s"Error on channel #${ctx.channel.id}", cause)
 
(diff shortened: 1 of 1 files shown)


driver/src/main/scala/core/protocol/MongoHandler.scala
```diff
@@ -94,6 +94,7 @@ private[reactivemongo] class MongoHandler(
   }
 
   override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable) = {
+    println(s"ERROR: $cause")
```
cchantep: There is a `log` just bellow

avdv: Yes, and I already told you that *if* one *does* enable the trace log level, that the exceptions would be printed. You, however, stated that "I haven't seen such case locally."

This println statement just makes it obvious that an exception is thrown almost for every single message flowing through the decoder.

I don't know what you want... :confused:

cchantep: If that's the reproducer, no need to create a PR, just ping me on a branch. That surprised me in this way.

avdv:
Closing, as this was never meant to be merged, I just wanted to trigger a Travis build.

PR #806, opened 2019-03-07
Use `System.nanoTime()` instead of `System.currentTimeMillis()`

avdv:
Using the latter for measuring elapsed times is not a good idea since it relates
to a fixed point in time and takes adjustments of the system clock into
account (clock skew, NTP updates) so time may seem "to be running backwards".

The `System.nanoTime()` clock does not have this problem, at least since Java 8,
and is monotonically increasing (when supported by the OS).

# Pull Request Checklist

## Background Context

I think this might be related to a lot of "[node] hasn't answered in time to last ping! Please check its connectivity" errors we are seeing.

According to the server logs / statistics the connections were steady.

## References

https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--

https://stackoverflow.com/questions/510462/is-system-nanotime-completely-useless/54566928#54566928

> Nonetheless, nanoTime() should still be preferred for implementing timed blocking, interval waiting, timeouts, etc. to currentTimeMillis() because the latter is a subject to the "time going backward" phenomenon (e. g. due to server time correction), i. e. currentTimeMillis() is not suitable for measuring time intervals at all.

https://stackoverflow.com/questions/510462/is-system-nanotime-completely-useless/54566928#54566928

> If you're just looking for extremely precise measurements of elapsed time, use System.nanoTime(). System.currentTimeMillis() will give you the most accurate possible elapsed time in milliseconds since the epoch, but System.nanoTime() gives you a nanosecond-precise time, relative to some arbitrary point.

http://vmnomad.blogspot.com/2015/09/esxi-and-guest-vm-time-sync-learning.html

### DIFF
--- a/driver/src/main/scala/core/protocol/MongoHandler.scala
+++ b/driver/src/main/scala/core/protocol/MongoHandler.scala
@@ -13,6 +13,7 @@ import reactivemongo.io.netty.handler.timeout.IdleStateEvent
 import reactivemongo.core.actors.{ ChannelConnected, ChannelDisconnected }
 
 import reactivemongo.util.LazyLogger
+import reactivemongo.util.timestamp
 
 private[reactivemongo] class MongoHandler(
   supervisor: String,
@@ -26,7 +27,7 @@ private[reactivemongo] class MongoHandler(
   override def channelActive(ctx: ChannelHandlerContext): Unit = {
     log(ctx, "Channel is active")
 
-    last = System.currentTimeMillis()
+    last = timestamp()
 
     receiver ! ChannelConnected(ctx.channel.id)
 
@@ -35,7 +36,7 @@ private[reactivemongo] class MongoHandler(
 
   override def channelIdle(ctx: ChannelHandlerContext, e: IdleStateEvent) = {
     if (last != -1L) {
-      val now = System.currentTimeMillis()
+      val now = timestamp()
 
       log(ctx, s"Channel has been inactive for ${now - last} (last = $last)")
     }
@@ -46,7 +47,7 @@ private[reactivemongo] class MongoHandler(
   }
 
   override def channelInactive(ctx: ChannelHandlerContext): Unit = {
-    val now = System.currentTimeMillis()
+    val now = timestamp()
 
     if (last != -1) {
       val chan = ctx.channel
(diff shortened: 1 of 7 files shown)


driver/src/main/scala/core/protocol/MongoHandler.scala
```diff
@@ -13,6 +13,7 @@ import reactivemongo.io.netty.handler.timeout.IdleStateEvent
 import reactivemongo.core.actors.{ ChannelConnected, ChannelDisconnected }
 
 import reactivemongo.util.LazyLogger
```
cchantep:
```suggestion
import reactivemongo.util.{ LazyLogger, timestamp }
```

avdv: Oh, I also removed the single import for timestamp and force-pushed again.

avdv:
Tests failed on Travis mainly because of the DNS resolution:
```
[error]   x resolve TXT record for gmail.com
[error]    Some(globalsign-smime-dv=CDYX+XFHUw2wml6/Gb8+59BsH31KzUr6c1l2BPvqKX8=) is Some but 'globalsign-smime-dv=CDYX+XFHUw2wml6/Gb8+59BsH31KzUr6c1l2BPvqKX8=' != 'v=spf1 redirect=_spf.google.com' 
```
I always see this error locally.

There's a single test (for the two builds) which also failed:
```
[info] Monitor should
[info]   + manage a single node DB system
[error]   ! manage unhandled Actor exception and Akka Restart
[error]    reactivemongo.core.actors.Exceptions$PrimaryUnavailableException: MongoError['No primary node is available! (Supervisor-1/monitorspec-con-881626603)'] (file:1)
[error] CAUSED BY
[error]  reactivemongo.core.actors.Exceptions$InternalState: null (file:1)
```
https://travis-ci.org/ReactiveMongo/ReactiveMongo/jobs/503208340

Have you seen this already?

avdv:
I was just thinking that it would be probably better to keep the nanosecond resolution instead of dividing by 1,000,000.

cchantep:
Then would prefer direct call to `nanoTime` w/o new internal until function

PR #808, opened 2019-03-10
Support $slice aggregation

cchantep:


driver/src/test/scala/Common.scala
```diff
@@ -58,6 +58,7 @@ object Common extends CommonAuth {
   val DefaultOptions = {
     val a = MongoConnectionOptions(
       failoverStrategy = failoverStrategy,
+      nbChannelsPerNode = 2, // TODO: Remove
```
mcintyre94: Should this have been merged?

PR #819, opened 2019-04-27
Refactor system history using fixed-size ring buffer

cchantep:


philipwhiuk:
Is this likely to / was this aimed at fixing a bug I've seen in 1.16.5 whereby you can accumulate large amounts of historical data:

historyBug

PR #862, opened 2019-08-23
Fixes bson macro for non sealed trait families

jtjeferreira:
# Pull Request Checklist

* [X] Have you added tests for any changed functionality?

## Fixes

Fixes bson macro for non sealed trait families, that was broken since https://github.com/ReactiveMongo/ReactiveMongo/pull/776

## Purpose

Make writing an non sealed trait family write the `className` field again.
The first commit demonstrates the problem via a testcase. The second commit fixes the issue.

## Background Context

I used the same approach used before. i.e use the `Macros.Options.UnionType[_]]` to decide if we should write the `className` field...

## References

https://github.com/ReactiveMongo/ReactiveMongo/pull/776

### DIFF
--- a/macros/src/test/scala/MacroSpec.scala
+++ b/macros/src/test/scala/MacroSpec.scala
@@ -200,6 +200,20 @@ final class MacroSpec extends org.specs2.mutable.Specification {
         } and roundtrip(a, format) and roundtrip(b, format)
     }
 
+    "handle union types in non sealed trait" in {
+      import Union._
+      import Macros.Options._
+      val a = UA2(1)
+      val b = UB2("hai")
+      val format = Macros.handlerOpts[UT2, UnionType[UA2 \/ UB2] with AutomaticMaterialization]
+
+      format.write(a).getAs[String]("className").
+        aka("class #1") must beSome("MacroTest.Union.UA2") and {
+          format.write(b).getAs[String]("className").
+            aka("class #2") must beSome("MacroTest.Union.UB2")
+        } and roundtrip(a, format) and roundtrip(b, format)
+    }
+
     "handle union types (ADT) with simple names" in {
       import Union._
       import Macros.Options._
--- a/macros/src/test/scala/MacroSpec.scala
+++ b/macros/src/test/scala/MacroSpec.scala
@@ -200,7 +200,7 @@ final class MacroSpec extends org.specs2.mutable.Specification {
         } and roundtrip(a, format) and roundtrip(b, format)
     }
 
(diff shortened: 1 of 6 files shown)


cchantep:
Thx for the contribution.

What's the goal of having a non-sealed family?
I would like to be sure that's a use case we want to support.

jtjeferreira:
> Thx for the contribution.
> 
> What's the goal of having a non-sealed family?
> I would like to be sure that's a use case we want to support.

Unfortunately I have legacy code which is a non-sealed trait... The `UnionType` is very useful for those cases, but it stopped working when we upgraded reactivemongo... Worst the macro accepts the non-sealed trait, but the `Writer` does not use the `className` field and I can no longer read the data from mongoDB

cchantep:
> Worst the macro accepts the non-sealed trait, but the Writer does not use the className field and I can no longer read the data from mongoDB ...

Don't understand ...

jtjeferreira:
> > Worst the macro accepts the non-sealed trait, but the Writer does not use the className field and I can no longer read the data from mongoDB ...
> 
> Don't understand ...

I meant that if you don't want to support non-sealed traits, would be nice that the macro rejects that code in compile time. The current problem is that the code is accepted, but it will not work in runtime...

cchantep:
Currently `Macros.handler[UT2]` is rejected at compile-time:

    No matching apply/unapply found: UT2

jtjeferreira:
> Currently `Macros.handler[UT2]` is rejected at compile-time:
> 
> ```
> No matching apply/unapply found: UT2
> ```

Yes, you are right. I was thinking about `Macros.handlerOpts[UT2, UnionType[UA2 \/ UB2]]` that compiles, but then fails at runtime (which is what this PR tries to fix)

jtjeferreira:
Hi @cchantep anything else I need to do here, so you can merge this?

macros/src/test/scala/MacroSpec.scala
```diff
@@ -200,6 +200,20 @@ final class MacroSpec extends org.specs2.mutable.Specification {
         } and roundtrip(a, format) and roundtrip(b, format)
     }
 
+    "handle union types in non sealed trait" in {
```
cchantep:
```suggestion
    "handle union types as sealed family" in {
```

jtjeferreira:
> Also need to be ported to https://github.com/reactivemongo/reactivemongo-bson

Will try to send a PR tomorrow

cchantep:
Thx

PR #876, opened 2019-09-28
Fix/fix bulk byte size

### LINKED ISSUE #871: Not all collection's items are added using "insert.many"
Hello,
there is an issue using insert.many. 
Mongo version:  4.0.6
Reactivemongo driver: 0.18.6

Code:
```
import reactivemongo.api.{MongoConnection, MongoDriver}
import reactivemongo.play.json.collection.JSONCollection
import play.api.libs.json.Json
import scala.concurrent.{ExecutionContext, Future}
import scala.util.Random

object GetStarted extends App {

  val stringSize = 1024

  // we expect 20000 documents will be inserted to test_col collection but
  // by running this script db.getCollection('test_col').count() 
  // we can see that just 3930 were added
  val collectionSize = 20000

  import ExecutionContext.Implicits.global

  val mongoUri = "mongodb://localhost:27017/mydb?authMode=scram-sha1"
  val driver = MongoDriver()
  val parsedUri = MongoConnection.parseURI(mongoUri)
  val connection = parsedUri.map(driver.connection)
  val futureConnection = Future.fromTry(connection)

  def db = futureConnection.flatMap(_.database("test_mongo40"))

  def testCollection =
    db.map(_.collection[JSONCollection]("test_col"))

  case class TestType(someData: String)

  implicit val testTypeFormat = Json.format[TestType]

  def writeTestData(testData: Seq[TestType]) =
    testCollection.flatMap(_.insert.many(testData))

  def makeItem = TestType(Random.alphanumeric.take(stringSize).mkString)

  val items = (1 to collectionSize).map(_ => makeItem)

  writeTestData(items)

}
```

build.sbt: 
```
name := "test_mongo40"

version := "0.1"

scalaVersion := "2.11.8"
val playVersion = "2.5.2"

libraryDependencies ++= Seq(
  "com.typesafe.play" %% "play" % playVersion,
  "org.reactivemongo" %% "play2-reactivemongo" % "0.18.6-play25",
  "org.apache.logging.log4j" % "log4j-core" % "2.12.1"
)
```

Issue:
3930 documents were created instead of 20000. 

Mongo logs:
```
E -        [conn608] Assertion: BSONObjectTooLarge: BSONObj size: 16878542 (0x1018BCE) is invalid. Size must be between 0 and 16793600(16MB) First element: insert: "test_col" src/mongo/bson/bsonobj.cpp 102
```

AurelienRichez:
# Pull Request Checklist

## Fixes

Probably fixes #871 

## Purpose

Correctly compute a bulk size in BulkOps. 

Currently we do not account for the total array size of a bulk. This means that a small errors adds up in our expected byte size for each document. This ultimately makes Reactive mongo send a bulk insert command which is too big.

## Background Context
When running a bulk operation, reactive mongo cuts the Seq of documents to insert (for instance) in bulks to avoid sending a command which is bigger than the maximum byte size allowed by MongoDB. But the way we compute the bulk size seems wrong. 

**Example:** 
Say we want to insert 3 empty documents: `[{},{},{}]` . Currently we compute the size by adding the size of an empty doc 3 times, but the real bulk size is the size of the complete array`[{},{},{}]`. 

This means that we can potentially send a command over the size limit.

A BSONArray is like a BSONDocument with numeric key according to the spec [1] : `[ {}, {}, {} ] <=> { '0': {}, '1': {}, '2': {}}`. This means that when computing the size of a bulk we must also add : 
 - The type prefix : 2 bytes
 - The size of the "key", which increase as the array gets bigger (The 11th element cost actually more bytes than the first because the key '10' take two bytes)

## References

[1] http://bsonspec.org/spec.html

### DIFF
--- a/bson/src/main/scala-2.13-/types.scala
+++ b/bson/src/main/scala-2.13-/types.scala
@@ -907,7 +907,7 @@ sealed trait BSONElementSet extends ElementProducer { self: BSONValue =>
   override private[reactivemongo] lazy val byteSize: Int =
     elements.foldLeft(5) {
       case (sz, BSONElement(n, v)) =>
-        sz + BSONElementSet.typePrefixByteSize + n.getBytes.size + v.byteSize
+        sz + BSONElementSet.docElementByteOverhead + n.getBytes.size + v.byteSize
     }
 }
 
@@ -919,7 +919,8 @@ object BSONElementSet {
   implicit def apply(set: Traversable[BSONElement]): BSONElementSet =
     new BSONDocument(set.map(Success(_)).toStream)
 
-  private[reactivemongo] val typePrefixByteSize = 2
+  // key null terminator + type prefix for the element
+  private[reactivemongo] val docElementByteOverhead = 2
 }
 
 /**
--- a/bson/src/main/scala-2.13-/types.scala
+++ b/bson/src/main/scala-2.13-/types.scala
@@ -907,7 +907,7 @@ sealed trait BSONElementSet extends ElementProducer { self: BSONValue =>
   override private[reactivemongo] lazy val byteSize: Int =
     elements.foldLeft(5) {
       case (sz, BSONElement(n, v)) =>
-        sz + 2 + n.getBytes.size + v.byteSize
+        sz + BSONElementSet.typePrefixByteSize + n.getBytes.size + v.byteSize
     }
 }
 
@@ -919,6 +919,7 @@ object BSONElementSet {
   implicit def apply(set: Traversable[BSONElement]): BSONElementSet =
     new BSONDocument(set.map(Success(_)).toStream)
 
+  private[reactivemongo] val typePrefixByteSize = 2
 }
 
 /**
--- a/driver/src/main/scala/api/collections/BulkOps.scala
+++ b/driver/src/main/scala/api/collections/BulkOps.scala
@@ -71,7 +71,7 @@ private[reactivemongo] object BulkOps {
           Left(s"size of document #${offset + docs} exceed the maxBsonSize: $bsz + 3 > $maxBsonSize")
         } else {
           val nc = docs + 1
-          val nsz = bsonSize + bsz
+          val nsz = bsonSize + bsz + docs.toString.getBytes.length + BSONElementSet.typePrefixByteSize
 
           if (nsz > maxBsonSize) {
             Right(BulkStage[I](
--- a/driver/src/main/scala/api/collections/BulkOps.scala
+++ b/driver/src/main/scala/api/collections/BulkOps.scala
@@ -68,7 +68,7 @@ private[reactivemongo] object BulkOps {
 
         // Total minimal size is key '1' size (1 byte) + type prefix (2 bytes)
         if (bsz + 1 + BSONElementSet.typePrefixByteSize > maxBsonSize) {
-          Left(s"size of document #${offset + docs} exceed the maxBsonSize: $bsz > $maxBsonSize")
+          Left(s"size of document #${offset + docs} exceed the maxBsonSize: $bsz + 3 > $maxBsonSize")
         } else {
           val nc = docs + 1
           val nsz = bsonSize + bsz
--- a/driver/src/main/scala/api/collections/BulkOps.scala
+++ b/driver/src/main/scala/api/collections/BulkOps.scala
@@ -2,6 +2,7 @@ package reactivemongo.api.collections
 
 import scala.concurrent.{ ExecutionContext, Future }
 
+import reactivemongo.bson.BSONElementSet
 import reactivemongo.core.errors.GenericDriverException
 
 /** Internal bulk operations */
@@ -66,7 +67,7 @@ private[reactivemongo] object BulkOps {
         val bsz = sz(doc)
 
         // Total minimal size is key '1' size (1 byte) + type prefix (2 bytes)
-        if (bsz + 1 + 2 > maxBsonSize) {
+        if (bsz + 1 + BSONElementSet.typePrefixByteSize > maxBsonSize) {
           Left(s"size of document #${offset + docs} exceed the maxBsonSize: $bsz > $maxBsonSize")
         } else {
           val nc = docs + 1
--- a/driver/src/main/scala/api/collections/BulkOps.scala
+++ b/driver/src/main/scala/api/collections/BulkOps.scala
@@ -65,7 +65,8 @@ private[reactivemongo] object BulkOps {
       case Some(doc) => {
         val bsz = sz(doc)
 
-        if (bsz > maxBsonSize) {
+        // Total minimal size is key '1' size (1 byte) + type prefix (2 bytes)
+        if (bsz + 1 + 2 > maxBsonSize) {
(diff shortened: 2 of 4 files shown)


driver/src/test/scala/BulkOpsSpec.scala
```diff
@@ -79,6 +77,34 @@ class BulkOpsSpec(implicit ee: ExecutionEnv)
...
+        maxBulkSize = 2)(_.byteSize) must beLike[BulkProducer[BSONDocument]] {
+        case prod1 =>
+          prod1() must beLeft
+      }
+    }
+
+    s"take into account the fact that keys increase in size in a larger array" in {
+      val ExpectedFirstBulk = Iterator.continually(doc1).take(10).toList
+      val ExpectedSecondBulk = List(doc1)
+      val enoughSizeWithoutConsideringSizeIncrease = (doc1.byteSize + 1 + BSONElementSet.typePrefixByteSize) * 11
```
cchantep: Let keep it in `maxBsonSize` (the value is used once)

driver/src/test/scala/BulkOpsSpec.scala
```diff
@@ -79,6 +77,34 @@ class BulkOpsSpec(implicit ee: ExecutionEnv)
...
+        documents = producer2Docs,
+        maxBsonSize = doc1.byteSize,
+        maxBulkSize = 2)(_.byteSize) must beLike[BulkProducer[BSONDocument]] {
+        case prod1 =>
+          prod1() must beLeft
+      }
+    }
+
+    s"take into account the fact that keys increase in size in a larger array" in {
+      val ExpectedFirstBulk = Iterator.continually(doc1).take(10).toList
```
cchantep:
```suggestion
      val ExpectedFirstBulk = List.fill(10)(doc1)
```

driver/src/test/scala/BulkOpsSpec.scala
```diff
@@ -79,6 +77,34 @@ class BulkOpsSpec(implicit ee: ExecutionEnv)
...
+      val ExpectedSecondBulk = List(doc1)
+      val enoughSizeWithoutConsideringSizeIncrease = (doc1.byteSize + 1 + BSONElementSet.typePrefixByteSize) * 11
+
+      bulks[BSONDocument](
+        documents = Iterator.continually(doc1).take(11).toSeq,
```
cchantep:
```suggestion
        documents = Seq.fill(11)(doc1),
```

driver/src/test/scala/BulkOpsSpec.scala
```diff
@@ -79,6 +77,34 @@ class BulkOpsSpec(implicit ee: ExecutionEnv)
...
+      bulks[BSONDocument](
+        documents = producer2Docs,
+        maxBsonSize = doc1.byteSize,
+        maxBulkSize = 2)(_.byteSize) must beLike[BulkProducer[BSONDocument]] {
+        case prod1 =>
+          prod1() must beLeft
+      }
+    }
+
+    s"take into account the fact that keys increase in size in a larger array" in {
```
cchantep:
```suggestion
    "take into account the fact that keys increase in size in a larger array" in {
```

driver/src/test/scala/BulkOpsSpec.scala
```diff
@@ -79,6 +77,34 @@ class BulkOpsSpec(implicit ee: ExecutionEnv)
         }
       }
     }
+
+    s"take minimal size into account into total size" in {
```
cchantep:
```suggestion
    "take minimal size into account into total size" in {
```

driver/src/test/scala/BulkOpsSpec.scala
```diff
@@ -79,6 +77,34 @@ class BulkOpsSpec(implicit ee: ExecutionEnv)
...
+      bulks[BSONDocument](
+        documents = producer2Docs,
+        maxBsonSize = doc1.byteSize,
+        maxBulkSize = 2)(_.byteSize) must beLike[BulkProducer[BSONDocument]] {
+        case prod1 =>
+          prod1() must beLeft
```
cchantep: Expected left value?

bson/src/main/scala-2.13-/types.scala
```diff
@@ -919,6 +919,7 @@ object BSONElementSet {
   implicit def apply(set: Traversable[BSONElement]): BSONElementSet =
     new BSONDocument(set.map(Success(_)).toStream)
 
+  private[reactivemongo] val typePrefixByteSize = 2
```
cchantep: As far as I can see/remind the [type code is written](https://github.com/ReactiveMongo/ReactiveMongo/blob/master/bson/src/main/scala/buffer/DefaultBufferHandler.scala#L81) as a single byte (or [there](https://github.com/mongodb/mongo-go-driver/blob/03f71cffc6c1b6c1de8b5c6d988d04916f4fbf7d/bson/bsonrw/value_reader.go#L684))

bson/src/main/scala-2.13-/types.scala
```diff
@@ -907,7 +907,7 @@ sealed trait BSONElementSet extends ElementProducer { self: BSONValue =>
   override private[reactivemongo] lazy val byteSize: Int =
```
cchantep: ` bson/src/main/scala-2.13+/types.scala` ?

driver/src/main/scala/api/collections/BulkOps.scala
```diff
@@ -65,11 +66,12 @@ private[reactivemongo] object BulkOps {
...
 
-        if (bsz > maxBsonSize) {
-          Left(s"size of document #${offset + docs} exceed the maxBsonSize: $bsz > $maxBsonSize")
+        // Total minimal size is key '1' size (1 byte) + type prefix (2 bytes)
+        if (bsz + 1 + BSONElementSet.typePrefixByteSize > maxBsonSize) {
+          Left(s"size of document #${offset + docs} exceed the maxBsonSize: $bsz + 3 > $maxBsonSize")
         } else {
           val nc = docs + 1
-          val nsz = bsonSize + bsz
+          val nsz = bsonSize + bsz + docs.toString.getBytes.length + BSONElementSet.typePrefixByteSize
```
cchantep:
```suggestion
          val keySize = docs.toString.getBytes.size // string repr. of current index used as key
          val nsz = bsonSize + bsz + keySize + BSONElementSet.typePrefixByteSize
```

Extract as `val` to comment the tricky part about key

bson/src/main/scala-2.13-/types.scala
AurelienRichez: Ah yes you're right ! In fact this counts the type code and the null character from the key I guess. I'll change the name

driver/src/test/scala/BulkOpsSpec.scala
AurelienRichez: Will do. This was to make it clear why I choose this max size.

bson/src/main/scala-2.13-/types.scala
AurelienRichez: I made the change in 2.13+

PR #945, opened 2020-02-26
Fix summoning of BSONHandler[Seq[Int]]

jtjeferreira:
## Fixes

Fixes implicit resolution of `BSONHandler[Seq[Int]]` due to a problem in `reactivemongo.bson.DefaultBSONHandlers#findWriter`

### DIFF
--- a/bson/src/main/scala/DefaultBSONHandlers.scala
+++ b/bson/src/main/scala/DefaultBSONHandlers.scala
@@ -159,7 +159,7 @@ trait DefaultBSONHandlers extends LowPrioBSONHandlers {
     def write(b: BSONJavaScript) = b
   }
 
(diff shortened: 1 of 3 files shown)


cchantep:
Thanks, rather check the new BSON API. This one will be remove after next major release.

jtjeferreira:
The 2nd commit has the fix...

jtjeferreira:
Hi @cchantep!

I am aware of the new API, and I think it is definitely an improvement over the current API. I am also aware of the compat layer that helps the migration...

The background of this fix is that I am working on using [magnolia](https://github.com/propensive/magnolia) to derive BSONWriters/BSONReaders/BSONHandlers, and this bug was preventing me to do it for BSONHandlers.
This bug is not a problem for the BSON API...

I will let you know when I have a more polished version, but meanwhile having this bug fixed would be awesome...

cchantep:
Effort based on the old API are not recommended.

PR #991, opened 2020-10-28
[WIP] Send the correct bulk of updates in BulkOps

maria-os:
# Pull Request Checklist

## Fixes

Fixes #xxxx

## Purpose

Send the correct bulk of updates in BulkOps.

## Background Context

Currently, we drop the first element of a bulk, assuming that it is passed as the first parameter to the `execute` method, however, what is passed as the first parameter is the first element of the whole sequence:
```
final def many(updates: Iterable[UpdateElement])(...) = updates.headOption match {
      case Some(first) => {
        val bulkProducer = BulkOps.bulks(
          updates, maxBsonSize, metadata.maxBulkSize) { up =>
          elementEnvelopeSize + pack.bsonSize(up.q) + pack.bsonSize(up.u)
        }

        BulkOps.bulkApply[UpdateElement, UpdateWriteResult](
          bulkProducer)({ bulk =>
          execute(first, bulk.drop(1).toSeq)                     // This is the problematic line 
        }, bulkRecover).map(MultiBulkWriteResult(_))
      }
```
### Example

Let's say, our updates are `[docA, docB, docC, docD]` and they're divided into 3 bulks `[[docA], [docB, docC], [docD]]`. Then code above will result in 3 calls to `execute`: `execute(docA, [])`, `execute(docA, [docC])`,  `execute(docA, [])`, i.e. `docB` and `docD` will never be written to mongo.

### Impact

1. We lose the first element of every bulk starting from the second bulk.
2. The bulk size is no longer correct because the first element of the whole updates sequence can be smaller or bigger than the first element of the bulk. 

> Why did you take this approach?

This is the simplest way to fix the issue. All other solutions / suggestions are welcomed.

## References

No

### DIFF
--- a/driver/src/main/scala/api/collections/UpdateOps.scala
+++ b/driver/src/main/scala/api/collections/UpdateOps.scala
@@ -179,7 +179,7 @@ trait UpdateOps[P <: SerializationPack] extends UpdateCommand[P]
 
         BulkOps.bulkApply[UpdateElement, UpdateWriteResult](
           bulkProducer)({ bulk =>
-          execute(first, bulk.drop(1).toSeq)
+          execute(bulk.headOption.getOrElse(first), bulk.drop(1).toSeq)
         }, bulkRecover).map(MultiBulkWriteResult(_))
       }
 
--- a/driver/src/test/scala/UpdateSpec.scala
+++ b/driver/src/test/scala/UpdateSpec.scala
@@ -10,7 +10,7 @@ import _root_.tests.Common
 
 import reactivemongo.api.tests.{ builder, decoder, pack, reader, writer }
 
-import scala.concurrent.{ Await, Future }
+import scala.concurrent.Future
 
 trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
   import reactivemongo.api.TestCompat._
@@ -76,7 +76,7 @@ trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
                 BSONDocument(
                   "id" -> BSONString(s"child${num}"),
                   "value" -> BSONString("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")))
-      
+
       def upsertElements(collection: DefaultCollection, documents: List[TestDocument]) = {
         val updateBuilder = collection.update(ordered = false)
         val updateElements = Future.sequence(
@@ -94,14 +94,14 @@ trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
 
       val childrenPerElement = 200 * 1000
       val documents = List(
-        TestDocument("id1", createChildDocuments(childrenPerElement)),
-        TestDocument("id2", createChildDocuments(childrenPerElement)))
+        TestDocument(System.nanoTime().toString, createChildDocuments(childrenPerElement)),
+        TestDocument(System.nanoTime().toString, createChildDocuments(childrenPerElement)))
 
-      val writeResult = Await.result(upsertElements(updCol2, documents), timeout)
+      val writeResult = upsertElements(updCol2, documents)
 
       val expectedDocumentsNumber = documents.size
-      writeResult.totalN must beTypedEqualTo(expectedDocumentsNumber) and {
-        writeResult.nModified must beTypedEqualTo(expectedDocumentsNumber)
+      writeResult.map(_.totalN) must beTypedEqualTo(expectedDocumentsNumber).await(1, timeout) and {
+        writeResult.map(_.nModified) must beTypedEqualTo(expectedDocumentsNumber).await(1, timeout)
       }
     }
 
--- a/driver/src/test/scala/UpdateSpec.scala
+++ b/driver/src/test/scala/UpdateSpec.scala
@@ -10,6 +10,8 @@ import _root_.tests.Common
 
 import reactivemongo.api.tests.{ builder, decoder, pack, reader, writer }
 
+import scala.concurrent.{ Await, Future }
+
 trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
   import reactivemongo.api.TestCompat._
   import Common._
@@ -60,6 +62,47 @@ trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
         }
       }
       section("gt_mongo32")
+    }
+
+    "update many documents with the default connection" in {
+      case class TestDocument(id: String, children: List[BSONDocument])
+
+      val createChildDocuments =
+        (childrenNumber: Int) =>
+          List
+            .range(0, childrenNumber)
+            .map(
+              num =>
+                BSONDocument(
+                  "id" -> BSONString(s"child${num}"),
+                  "value" -> BSONString("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")))
+      
+      def upsertElements(collection: DefaultCollection, documents: List[TestDocument]) = {
+        val updateBuilder = collection.update(ordered = false)
+        val updateElements = Future.sequence(
+          for {
+            document <- documents
+          } yield updateBuilder.element(
+            q = BSONDocument("_id" -> document.id),
+            u = BSONDocument(
+              "_id" -> document.id,
+              "children" -> BSONArray(document.children)),
+            upsert = true))
+
+        updateElements.flatMap(updates => updateBuilder.many(updates))
+      }
+
+      val childrenPerElement = 200 * 1000
+      val documents = List(
+        TestDocument("id1", createChildDocuments(childrenPerElement)),
+        TestDocument("id2", createChildDocuments(childrenPerElement)))
+
+      val writeResult = Await.result(upsertElements(updCol2, documents), timeout)
+
+      val expectedDocumentsNumber = documents.size
+      writeResult.totalN must beTypedEqualTo(expectedDocumentsNumber) and {
+        writeResult.nModified must beTypedEqualTo(expectedDocumentsNumber)
+      }
     }
 
     "upsert a document" >> {


driver/src/main/scala/api/collections/UpdateOps.scala
```diff
@@ -170,22 +170,23 @@ trait UpdateOps[P <: SerializationPack] extends UpdateCommand[P]
...
-          updates, maxBsonSize, metadata.maxBulkSize) { up =>
-          elementEnvelopeSize + pack.bsonSize(up.q) + pack.bsonSize(up.u)
+          updates, maxBsonSize, metadata.maxBulkSize) {
+          up =>
+            elementEnvelopeSize + pack.bsonSize(up.q) + pack.bsonSize(up.u)
         }
 
         BulkOps.bulkApply[UpdateElement, UpdateWriteResult](
-          bulkProducer)({ bulk =>
-          execute(first, bulk.drop(1).toSeq)
```
maria-os, line 182: Here is the bug: we override the first element of the bulk by the first element of the sequence.

driver/src/test/scala/UpdateSpec.scala
```diff
@@ -62,6 +59,55 @@ trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
...
+
+      def spec(c: DefaultCollection, timeout: FiniteDuration) = {
+        val childrenPerElement = 200 * 1000
+        val elements = List(
+          TestDocument("id1", mockChildElements(childrenPerElement)),
+          TestDocument("id2", mockChildElements(childrenPerElement)))
+
+        val writeResult = Await.result(upsertElements(c, elements), timeout)
+        // Wait to make sure the operation is fully finished.
+        Thread.sleep(1000L)
```
maria-os, line 99: This is not really needed. Added only for a demonstration.

driver/src/test/scala/UpdateSpec.scala
```diff
@@ -62,6 +59,55 @@ trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
...
+            q = BSONDocument("id" -> element.id),
+            u = BSONDocument(
+              "id" -> element.id,
+              "children" -> BSONArray(element.children)),
+            upsert = true))
+
+        updateElements.flatMap(updates => updateBuilder.many(updates))
+      }
+
+      def spec(c: DefaultCollection, timeout: FiniteDuration) = {
```
cchantep, line 91: No need for a separate function, directly in `in` block

driver/src/test/scala/UpdateSpec.scala
cchantep: So remove it

driver/src/test/scala/UpdateSpec.scala
```diff
@@ -62,6 +59,55 @@ trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
...
+        val elements = List(
+          TestDocument("id1", mockChildElements(childrenPerElement)),
+          TestDocument("id2", mockChildElements(childrenPerElement)))
+
+        val writeResult = Await.result(upsertElements(c, elements), timeout)
+        // Wait to make sure the operation is fully finished.
+        Thread.sleep(1000L)
+
+        val expectedDocumentsNumber = elements.size
+        writeResult.totalN must beEqualTo(expectedDocumentsNumber)
```
cchantep, line 102:
```suggestion
        writeResult.totalN must beTypedEqualTo(expectedDocumentsNumber) and {
```

driver/src/main/scala/api/collections/UpdateOps.scala
```diff
@@ -139,7 +139,7 @@ trait UpdateOps[P <: SerializationPack] extends UpdateCommand[P]
 
       BulkOps.bulkApply[UpdateElement, UpdateWriteResult](
         bulkProducer)(
-        { bulk => execute(firstUpdate, bulk.toSeq) },
```
cchantep, line 142: This break the safety to execute with at least on element. Not acceptable.

driver/src/main/scala/api/collections/UpdateOps.scala
```diff
@@ -170,22 +170,23 @@ trait UpdateOps[P <: SerializationPack] extends UpdateCommand[P]
...
      * }}}
      */
-    final def many(updates: Iterable[UpdateElement])(implicit ec: ExecutionContext): Future[MultiBulkWriteResult] = updates.headOption match {
-      case Some(first) => {
+    final def many(updates: Iterable[UpdateElement])(implicit ec: ExecutionContext): Future[MultiBulkWriteResult] = {
+      if (updates.nonEmpty) {
         val bulkProducer = BulkOps.bulks(
-          updates, maxBsonSize, metadata.maxBulkSize) { up =>
-          elementEnvelopeSize + pack.bsonSize(up.q) + pack.bsonSize(up.u)
+          updates, maxBsonSize, metadata.maxBulkSize) {
```
cchantep, line 176: Please do not break format. Revert that.

driver/src/main/scala/api/collections/UpdateOps.scala
cchantep: The `first` element to `updates` is the first to be executed as bulk.
Do not see your point.

cchantep (commented):
For now neither the issue is clear, nor the changes could be anyway accepted

driver/src/test/scala/UpdateSpec.scala
```diff
@@ -62,6 +59,55 @@ trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
...
+        updateElements.flatMap(updates => updateBuilder.many(updates))
+      }
+
+      def spec(c: DefaultCollection, timeout: FiniteDuration) = {
+        val childrenPerElement = 200 * 1000
+        val elements = List(
+          TestDocument("id1", mockChildElements(childrenPerElement)),
+          TestDocument("id2", mockChildElements(childrenPerElement)))
+
+        val writeResult = Await.result(upsertElements(c, elements), timeout)
```
cchantep, line 97: Rather use specs2 `.await`

driver/src/main/scala/api/collections/UpdateOps.scala
```diff
@@ -194,20 +195,11 @@ trait UpdateOps[P <: SerializationPack] extends UpdateCommand[P]
...
-        upsert = false,
-        multi = false,
-        collation = None,
-        arrayFilters = Seq.empty)
-
       // Command envelope to compute accurate BSON size limit
       val emptyCmd = new ResolvedCollectionCommand(
         collection.name,
         new Update(
-          emptyElm, Seq.empty, ordered, writeConcern, false))
```
cchantep, line 210: Doesn't make sense to me.
Minimal bulk size should contain at least one empty document.

driver/src/main/scala/api/collections/UpdateOps.scala
maria-os: The point is the first element of the second bulk of updates will never be written to mongo, the same for the third bulk etc. The code `drops` the first element of bulks, assuming that the first element is passed as the first argument to `execute`, which is not true.
If the explanation isn't clear yet, please see the added test. The test creates 2 elements of equal size, writes them to mongo and checks the number of modified elements. The elements are so big that they need to be written in 2 bulks. Without the fix, i.e. the second commit, the test fails.

driver/src/test/scala/UpdateSpec.scala
maria-os: Done.

driver/src/main/scala/api/collections/UpdateOps.scala
```diff
@@ -19,8 +19,8 @@ import reactivemongo.api.commands.{
 }
 
 /**
- * @define writeConcernParam the [[https://docs.mongodb.com/manual/reference/write-concern/ writer concern]] to be used
- * @define orderedParam the [[https://docs.mongodb.com/manual/reference/method/db.collection.update/#perform-an-unordered-update ordered]] behaviour
+ * @define writeConcernParam             the [[https://docs.mongodb.com/manual/reference/write-concern/ writer concern]] to be used
```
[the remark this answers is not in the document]
maria-os: Done.

driver/src/main/scala/api/collections/UpdateOps.scala
```diff
@@ -32,8 +32,8 @@ trait UpdateOps[P <: SerializationPack] extends UpdateCommand[P]
     collection.db.connectionState.metadata.maxWireVersion
 
   /**
-   * @param ordered $orderedParam
-   * @param writeConcern $writeConcernParam
+   * @param ordered                  $orderedParam
```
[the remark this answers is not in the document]
maria-os: Done.

driver/src/test/scala/UpdateSpec.scala
```diff
@@ -1,15 +1,12 @@
 import scala.concurrent.duration.FiniteDuration
-
```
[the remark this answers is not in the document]
maria-os: Done.

driver/src/test/scala/UpdateSpec.scala
```diff
@@ -62,6 +59,55 @@ trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
       section("gt_mongo32")
     }
 
+    "update many documents" >> {
```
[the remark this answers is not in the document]
maria-os: Done but not sure I understood the comment. Please, be more specific.

driver/src/test/scala/UpdateSpec.scala
```diff
@@ -62,6 +59,55 @@ trait UpdateSpec extends UpdateFixtures { collectionSpec: CollectionSpec =>
...
+      case class TestDocument(id: String, children: List[BSONDocument])
+
+      def mockChildElements =
```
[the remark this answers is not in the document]
maria-os: Fixed. Regarding the name - I'm open for your suggestions.

driver/src/test/scala/UpdateSpec.scala
maria-os: Fixed, I suppose.

driver/src/main/scala/api/collections/UpdateOps.scala
maria-os: Done.

driver/src/test/scala/UpdateSpec.scala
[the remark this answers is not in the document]
cchantep: `>>` is only to group multiple `in`

driver/src/test/scala/UpdateSpec.scala
maria-os: Done, I guess.

driver/src/main/scala/api/collections/UpdateOps.scala
maria-os: Reverted. Updated the fix to minimise the changes.

maria-os:
Hi [Cédric](@cchantep cchantep),

Thank you for the quick reply. We opened this pull request as a WIP to trigger the integration tests and to check ourselves. We were planning first to open a thread on the [mailing list](https://groups.google.com/forum/#!forum/reactivemongo) as the guidelines suggest but since you have already answered we will be happy to skip this step.

As you may guess we think there is a bug in `BulkOps`. In a few words, the many method breaks bulks by passing wrong parameters to execute. It drops the first element of every bulk and passes the first element of updates, which causes the actual bulk operation to run on wrong elements.

Please, see the pull request description for details and the test reproducing the issue.

Thank you,
Dor and Maria

driver/src/main/scala/api/collections/UpdateOps.scala
maria-os: Reverted. We change the original fix to minimise the diff.

cchantep:
Thanks for the report; Fix in another pull request.

maria-os:
Thank you for the quick fix!

It would be great if you release a new version or hotfix 0.20.x as currently we cannot upgrade ReactiveMongo due to this bug.

Thank you again for your time and for this project!

cchantep:
https://search.maven.org/artifact/org.reactivemongo/reactivemongo_2.12/0.20.13/jar

PR #1097, opened 2022-01-07
Closes #1096

### LINKED ISSUE #1096: Duplicated `deletes` elements in serialized `DeleteCmd`
### ReactiveMongo Version (0.10.5, 0.11.6 / etc)

`1.0.10`

### MongoDB version (2.6, 3.0 / etc)

`5.0.3`, but it does not depend on the specific MongoDB version.

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)

macOS Big Sur Version `11.6`

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

OpenJDK 17.0.1

### Expected Behavior

```scala
collection.delete(ordered = true).one(q = BSONDocument("_id" -> 1), limit = Some(1))
```

must be serialized as:

```bson
{
  'delete': 'foo',
  'ordered': true,
  'writeConcern': {
    'w': 1,
    'j': false
  },
  'deletes': [
    {
      'q': {
        '_id': 1
      },
      'limit': 1
    }
  ]
}
```

### Actual Behavior

During serialization of `DeleteCmd` the first element in `deletes` collection duplicated in resulted BSON. 
For example: 

```scala
collection.delete(ordered = true).one(q = BSONDocument("_id" -> 1), limit = Some(1))
```

will be serialized as:

```bson
{
  'delete': 'foo',
  'ordered': true,
  'writeConcern': {
    'w': 1,
    'j': false
  },
  'deletes': [
    {
      'q': {
        '_id': 1
      },
      'limit': 1
    },
    {
      'q': {
        '_id': 1
      },
      'limit': 1
    }
  ]
}
```

Because of this `delete.one` currently does not match the semantics of the original `deleteOne()` operations and deletes at least 2 documents, but not 1.

### Solution

#1097

Romastyi:
# Pull Request Checklist

## Fixes

Fixes #1096 

## Purpose

This is a fix of serialization for `DeleteCmd` because the current implementation is incorrect in some cases described in #1096.

## Background Context

Need `delete.one()` to match the original semantics of `deleteOne()`.

### DIFF
--- a/driver/src/main/scala/api/commands/DeleteCommand.scala
+++ b/driver/src/main/scala/api/commands/DeleteCommand.scala
@@ -13,7 +13,7 @@ import reactivemongo.api.{
  */
 private[reactivemongo] trait DeleteCommand[P <: SerializationPack] { self: PackSupport[P] =>
 
-  private[api] final class Delete(
+  private[reactivemongo] final class Delete(
     val deletes: Seq[DeleteElement],
     val ordered: Boolean,
     val writeConcern: WriteConcern) extends CollectionCommand with CommandWithResult[DeleteResult] {
@@ -21,7 +21,7 @@ private[reactivemongo] trait DeleteCommand[P <: SerializationPack] { self: PackS
   }
 
   /** Delete command element */
-  final class DeleteElement private[api] (
+  final class DeleteElement private[reactivemongo] (
     _q: pack.Document,
     _limit: Int,
     _collation: Option[Collation]) {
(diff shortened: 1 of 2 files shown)


driver/src/main/scala/api/commands/DeleteCommand.scala
```diff
@@ -13,15 +13,15 @@ import reactivemongo.api.{
  */
 private[reactivemongo] trait DeleteCommand[P <: SerializationPack] { self: PackSupport[P] =>
 
-  private[api] final class Delete(
+  private[reactivemongo] final class Delete(
```
cchantep, line 16: Revert

driver/src/main/scala/api/commands/DeleteCommand.scala
```diff
@@ -84,7 +84,7 @@ private[reactivemongo] trait DeleteCommand[P <: SerializationPack] { self: PackS
       delete.command.deletes.headOption.foreach { first =>
         elements += element("deletes", builder.array(
           writeElement(builder, first) +:
-            delete.command.deletes.map(writeElement(builder, _))))
+            delete.command.deletes.tail.map(writeElement(builder, _))))
```
cchantep, line 87: ✅

cchantep (changes requested):

driver/src/main/scala/api/commands/DeleteCommand.scala
Romastyi: It's needed for tests implementation. See `driver/src/test/scala/DeleteCommandSpec.scala:52`. 

This implementation was done in analogue with already existed `driver/src/test/scala/UpdateCommandSpec.scala`. In this spec `UpdateCommand#Update` is used and this is also defined as `private[reactivemongo]`.

If it is not a good pattern for testing, @cchantep pls, advise another solution.

PR #1189, opened 2022-12-13
Add MongoConnectionOptions.nettyEventLoopThreads

ornicar:
Allows configuring the number of threads in the netty event loop. Defaults to zero, which lets netty decide how many threads to spawn. Netty will by default select 2 times the number of available processors. That's what it already was doing before adding this setting, so the default behaviour remains the same.

[Similar to, and inspired from, `play.server.netty.eventLoopThreads`](https://github.com/playframework/playframework/blob/0ac9a28c19491acb796ce6d460c848efc201e789/transport/server/play-netty-server/src/main/resources/reference.conf#L3-L9).

Use case: we run lila on a machine with 96 processors, and reactivemongo doesn't need 192 netty threads.


cchantep:
https://app.circleci.com/pipelines/github/ReactiveMongo/ReactiveMongo/1332/workflows/1965d621-88c3-4759-8c1c-b9ba6f347983/jobs/17661?invite=true#step-112-308

PR #1245, opened 2023-08-06
Publish pekko based ReactiveMongo binaries 

### LINKED ISSUE #1244: Pekko support
### Feature request

[Pekko](https://pekko.apache.org/what-is-pekko.html), the OSS fork of Akka has reached 1.0.0. Many projects will want to move away from Akka since their introduction of the Business Source License model for Akka version 2.7 and above.
A pekko based distribution would allow people to keep using ReactiveMongo.

SamTheisens:
# Pull Request Checklist

## Fixes

Fixes #1244

## Purpose

Create a pekko based ReactiveMongo distribution for those who want to adopt Pekko.

## Background Context

Implementation details
 - Introduce neutral aliases for Akka actor dependencies to minimize duplication
 - Pulls in akka or pekko dependencies based on a build parameter

Features
- Publishes pekko variants of all modules
- Adds pekko based permutations to integration test matrix
- Does not build scala 3 version for pekko yet, because Pekko needs Scala 3.3.0 or higher 

Some integration tests are failing intermittently with timeouts, but it doesn't seem related to the changes.

Related PR on documentation repo https://github.com/ReactiveMongo/reactivemongo-site/pull/316

### DIFF
--- a/actors-akka/src/main/scala/reactivemongo/actors/actor/package.scala
+++ b/actors-akka/src/main/scala/reactivemongo/actors/actor/package.scala
@@ -1,0 +1,16 @@
+package reactivemongo.actors
+
+package object actor {
+  type ActorSystem = akka.actor.ActorSystem
+  val ActorSystem = akka.actor.ActorSystem
+
+  type ActorRef = akka.actor.ActorRef
+  type Actor = akka.actor.Actor
+  val Actor = akka.actor.Actor
+
+  type Props = akka.actor.Props
+  val Props = akka.actor.Props
+  type Terminated = akka.actor.Terminated
+  type Cancellable = akka.actor.Cancellable
+
+}
--- a/actors-akka/src/main/scala/reactivemongo/actors/actor/package.scala
+++ b/actors-akka/src/main/scala/reactivemongo/actors/actor/package.scala
@@ -12,5 +12,4 @@ package object actor {
   val Props = akka.actor.Props
   type Terminated = akka.actor.Terminated
   type Cancellable = akka.actor.Cancellable
-
 }
--- a/actors-akka/src/main/scala/reactivemongo/actors/util/package.scala
+++ b/actors-akka/src/main/scala/reactivemongo/actors/util/package.scala
@@ -1,0 +1,15 @@
+package reactivemongo.actors
+
+import akka.util.{
+  ByteString => PekkoByteString,
+  Timeout => PekkoTimeout
+}
+
+package object util {
+  type Timeout = PekkoTimeout
+  val Timeout = PekkoTimeout
+
+  type ByteString = PekkoByteString
+  val ByteString = PekkoByteString
+
+}
--- a/actors-akka/src/main/scala/reactivemongo/actors/util/package.scala
+++ b/actors-akka/src/main/scala/reactivemongo/actors/util/package.scala
@@ -1,15 +1,9 @@
 package reactivemongo.actors
 
-import akka.util.{
-  ByteString => PekkoByteString,
-  Timeout => PekkoTimeout
-}
-
 package object util {
-  type Timeout = PekkoTimeout
-  val Timeout = PekkoTimeout
+  type Timeout = akka.util.Timeout
+  val Timeout = akka.util.Timeout
 
-  type ByteString = PekkoByteString
-  val ByteString = PekkoByteString
-
+  type ByteString = akka.util.ByteString
+  val ByteString = akka.util.ByteString
 }
--- a/actors-akka/src/test/scala/reactivemongo/actors/testkit/package.scala
+++ b/actors-akka/src/test/scala/reactivemongo/actors/testkit/package.scala
@@ -1,9 +1,8 @@
 package reactivemongo.actors
 
 import akka.actor.Actor
-import akka.testkit.{ TestActorRef => PekkoTestActorRef }
 
 package object testkit {
-  type TestActorRef[T <: Actor] = PekkoTestActorRef[T]
-  val TestActorRef = PekkoTestActorRef
+  type TestActorRef[T <: Actor] = akka.testkit.TestActorRef[T]
+  val TestActorRef = akka.testkit.TestActorRef
 }
--- a/actors-akka/src/test/scala/reactivemongo/actors/testkit/package.scala
+++ b/actors-akka/src/test/scala/reactivemongo/actors/testkit/package.scala
@@ -1,0 +1,9 @@
+package reactivemongo.actors
+
+import akka.actor.Actor
+import akka.testkit.{ TestActorRef => PekkoTestActorRef }
+
+package object testkit {
+  type TestActorRef[T <: Actor] = PekkoTestActorRef[T]
+  val TestActorRef = PekkoTestActorRef
+}
--- a/actors-pekko/src/main/scala/reactivemongo/actors/util/package.scala
+++ b/actors-pekko/src/main/scala/reactivemongo/actors/util/package.scala
@@ -1,0 +1,15 @@
+package reactivemongo.actors
+
+import org.apache.pekko.util.{
+  ByteString => PekkoByteString,
+  Timeout => PekkoTimeout
+}
+
+package object util {
+  type Timeout = PekkoTimeout
+  val Timeout = PekkoTimeout
+
+  type ByteString = PekkoByteString
+  val ByteString = PekkoByteString
+
+}
--- a/actors-pekko/src/main/scala/reactivemongo/actors/util/package.scala
+++ b/actors-pekko/src/main/scala/reactivemongo/actors/util/package.scala
@@ -1,15 +1,9 @@
 package reactivemongo.actors
 
-import org.apache.pekko.util.{
-  ByteString => PekkoByteString,
-  Timeout => PekkoTimeout
-}
-
 package object util {
-  type Timeout = PekkoTimeout
-  val Timeout = PekkoTimeout
+  type Timeout = org.apache.pekko.util.Timeout
+  val Timeout = org.apache.pekko.util.Timeout
 
-  type ByteString = PekkoByteString
-  val ByteString = PekkoByteString
-
+  type ByteString = org.apache.pekko.util.ByteString
+  val ByteString = org.apache.pekko.util.ByteString
 }
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,6 @@
+import sbt.*
+import sbt.Keys.*
+
 lazy val `ReactiveMongo-Core` = project
   .in(file("core"))
   .settings(
@@ -51,14 +54,28 @@ lazy val `ReactiveMongo-Core` = project
     }
   )
 
-lazy val `ReactiveMongo` = new Driver(`ReactiveMongo-Core`).module
+
+lazy val `Actors-Akka` = project.in(file("actors-akka")).settings(
+  libraryDependencies ++= Dependencies.akka.value
+)
+
+lazy val `Actors-Pekko` = project.in(file("actors-pekko")).settings(
+  libraryDependencies ++= Dependencies.pekko.value
+)
+
+lazy val actorModule = Common.actorModule match {
+  case "pekko" => `Actors-Pekko`
+  case "akka" => `Actors-Akka`
+}
+
+lazy val `ReactiveMongo` = new Driver(`ReactiveMongo-Core`, actorModule).module
 
 lazy val `ReactiveMongo-Test` = project
   .in(file("test"))
   .settings(
     description := "ReactiveMongo test helpers"
   )
-  .dependsOn(`ReactiveMongo`)
+  .dependsOn(`ReactiveMongo`).dependsOn()
 
 // ---
 
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,3 @@
-import sbt.*
-import sbt.Keys.*
-
 lazy val `ReactiveMongo-Core` = project
   .in(file("core"))
   .settings(
--- a/build.sbt
+++ b/build.sbt
@@ -55,17 +55,17 @@ lazy val `ReactiveMongo-Core` = project
   )
 
 
-lazy val `Actors-Akka` = project.in(file("actors-akka")).settings(
+lazy val `ReactiveMongo-Actors-Akka` = project.in(file("actors-akka")).settings(
   libraryDependencies ++= Dependencies.akka.value
 )
 
-lazy val `Actors-Pekko` = project.in(file("actors-pekko")).settings(
+lazy val `ReactiveMongo-Actors-Pekko` = project.in(file("actors-pekko")).settings(
   libraryDependencies ++= Dependencies.pekko.value
 )
 
 lazy val actorModule = Common.actorModule match {
-  case "pekko" => `Actors-Pekko`
-  case "akka" => `Actors-Akka`
+  case "pekko" => `ReactiveMongo-Actors-Pekko`
+  case "akka" => `ReactiveMongo-Actors-Akka`
 }
 
 lazy val `ReactiveMongo` = new Driver(`ReactiveMongo-Core`, actorModule).module
--- a/build.sbt
+++ b/build.sbt
@@ -75,7 +75,7 @@ lazy val `ReactiveMongo-Test` = project
   .settings(
     description := "ReactiveMongo test helpers"
   )
-  .dependsOn(`ReactiveMongo`).dependsOn()
+  .dependsOn(`ReactiveMongo`)
 
 // ---
 
--- a/driver/src/main/scala/api/AsyncDriver.scala
+++ b/driver/src/main/scala/api/AsyncDriver.scala
@@ -18,11 +18,11 @@ import reactivemongo.core.actors.{
   StandardDBSystemWithX509
 }
 import reactivemongo.core.nodeset.Authenticate
-
-import akka.actor.{ Actor, ActorRef, ActorSystem, Props, Terminated }
-import akka.pattern.ask
-import akka.util.Timeout
+import reactivemongo.actors.actor.{ Actor, ActorRef, ActorSystem }
+import reactivemongo.actors.pattern.ask._
+import reactivemongo.actors.util.Timeout
 import com.typesafe.config.Config
+import reactivemongo.actors.actor.{ Props, Terminated }
 
 /**
  * The asynchronous driver (see [[MongoConnection]]).
@@ -38,7 +38,6 @@ import com.typesafe.config.Config
  *
  * @param config a custom configuration (otherwise the default options are used)
  * @param classLoader a classloader used to load the actor system
- *
  * @define parsedURIParam the URI parsed by [[reactivemongo.api.MongoConnection.fromString]]
  * @define connectionNameParam the name for the connection pool
  * @define optionsParam the options for the new connection pool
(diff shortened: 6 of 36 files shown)


project/Common.scala
```diff
@@ -51,15 +57,15 @@ object Common extends AutoPlugin {
     "Use ReactiveMongo-Shaded (see system property 'reactivemongo.shaded')"
   )
 
+  lazy val scalaVersions: Seq[String] = {
+    if (actorModule == "akka") Seq(scala211, scala212, scala213, scala31)
+    else Seq(scala212, scala213)
```
pjfanning, line 62: pekko supports scala3 - but only scala3.3 and above

pjfanning: Common.scala31 is already set to 3.2.2. Maybe, it should be renamed to scala3 and set to 3.3.0. That would allow pekko support for ReactiveMongo support to also include scala3.

SamTheisens: Indeed. But at this point ReactiveMongo builds against Scala `3.2.2` https://github.com/SamTheisens/ReactiveMongo/blob/master/project/Common.scala#L34
Not sure there's anything against upgrading to `3.3.x`, but perhaps that needs to be done in a separate PR?

actors-akka/src/main/scala/reactivemongo/actors/actor/package.scala
```diff
@@ -0,0 +1,16 @@
...
+
+  type ActorRef = akka.actor.ActorRef
+  type Actor = akka.actor.Actor
+  val Actor = akka.actor.Actor
+
+  type Props = akka.actor.Props
+  val Props = akka.actor.Props
+  type Terminated = akka.actor.Terminated
+  type Cancellable = akka.actor.Cancellable
+
```
cchantep, line 15:
```suggestion
```

actors-akka/src/main/scala/reactivemongo/actors/util/package.scala
```diff
@@ -0,0 +1,15 @@
...
+  Timeout => PekkoTimeout
+}
+
+package object util {
+  type Timeout = PekkoTimeout
+  val Timeout = PekkoTimeout
+
+  type ByteString = PekkoByteString
+  val ByteString = PekkoByteString
+
```
cchantep, line 14:
```suggestion
```

actors-akka/src/main/scala/reactivemongo/actors/util/package.scala
```diff
@@ -0,0 +1,15 @@
+package reactivemongo.actors
+
+import akka.util.{
+  ByteString => PekkoByteString,
```
cchantep, line 4: Don't see why Akka type should be prefixed as `Pekko`

actors-pekko/src/main/scala/reactivemongo/actors/util/package.scala
```diff
@@ -0,0 +1,15 @@
...
+  Timeout => PekkoTimeout
+}
+
+package object util {
+  type Timeout = PekkoTimeout
+  val Timeout = PekkoTimeout
+
+  type ByteString = PekkoByteString
+  val ByteString = PekkoByteString
+
```
cchantep, line 14:
```suggestion
```

build.sbt
```diff
@@ -51,14 +54,28 @@ lazy val `ReactiveMongo-Core` = project
     }
   )
 
-lazy val `ReactiveMongo` = new Driver(`ReactiveMongo-Core`).module
+
+lazy val `Actors-Akka` = project.in(file("actors-akka")).settings(
```
cchantep, line 58: SBT module naming

build.sbt
```diff
@@ -51,14 +54,28 @@ lazy val `ReactiveMongo-Core` = project
...
+
+lazy val `ReactiveMongo` = new Driver(`ReactiveMongo-Core`, actorModule).module
 
 lazy val `ReactiveMongo-Test` = project
   .in(file("test"))
   .settings(
     description := "ReactiveMongo test helpers"
   )
-  .dependsOn(`ReactiveMongo`)
+  .dependsOn(`ReactiveMongo`).dependsOn()
```
cchantep, line 78: !!

driver/src/main/scala/api/AsyncDriver.scala
```diff
@@ -38,7 +38,6 @@ import com.typesafe.config.Config
  *
  * @param config a custom configuration (otherwise the default options are used)
  * @param classLoader a classloader used to load the actor system
- *
```
cchantep, line 41: revert

driver/src/main/scala/api/AsyncDriver.scala
```diff
@@ -18,11 +18,11 @@ import reactivemongo.core.actors.{
   StandardDBSystemWithX509
 }
 import reactivemongo.core.nodeset.Authenticate
-
-import akka.actor.{ Actor, ActorRef, ActorSystem, Props, Terminated }
-import akka.pattern.ask
-import akka.util.Timeout
+import reactivemongo.actors.actor.{ Actor, ActorRef, ActorSystem }
```
cchantep, line 21: import order (`scalafix`)

driver/src/main/scala/api/Failover.scala
```diff
@@ -1,12 +1,11 @@
 package reactivemongo.api
 
+import reactivemongo.actors.pattern.after
```
cchantep, line 3: import order

project/Dependencies.scala
```diff
@@ -31,6 +31,15 @@ object Dependencies {
     )
   }
 
+  val pekko = Def.setting[Seq[ModuleID]] {
+    val ver = "1.0.1"
+    Seq(
```
cchantep, line 36:
```suggestion

    Seq(
```

project/Driver.scala
```diff
@@ -150,6 +152,7 @@ final class Driver(core: Project) {
       }
     }
     .dependsOn(core)
+    .dependsOn(actorModule % "compile->compile;test->test")
```
cchantep, line 155: As far as I see, there is no test in this "binding" module, so the scopes there are not required

cchantep (changes requested):
Thx for the contributions. Few comments.

project/Driver.scala
SamTheisens: No tests, but there is code for the test scope in there: https://github.com/ReactiveMongo/ReactiveMongo/pull/1245/files#diff-a2cdc7f0cb0342178b80349f37ccab46ca61781e5cc29e56e5b156099620eb67

actors-akka/src/main/scala/reactivemongo/actors/actor/package.scala
SamTheisens: Addressed in 87d3301898f6bbf86f7167a2c141cc2bbd9e8370

actors-akka/src/main/scala/reactivemongo/actors/util/package.scala
SamTheisens: Addressed in 87d3301898f6bbf86f7167a2c141cc2bbd9e8370

actors-pekko/src/main/scala/reactivemongo/actors/util/package.scala
SamTheisens: Addressed in 87d3301898f6bbf86f7167a2c141cc2bbd9e8370

build.sbt
SamTheisens: What would you like the module name to be?

build.sbt
SamTheisens: Addressed in 87d3301898f6bbf86f7167a2c141cc2bbd9e8370

driver/src/main/scala/api/AsyncDriver.scala
SamTheisens: Addressed in 87d3301898f6bbf86f7167a2c141cc2bbd9e8370

driver/src/main/scala/api/AsyncDriver.scala
SamTheisens: Addressed in 87d3301898f6bbf86f7167a2c141cc2bbd9e8370

project/Dependencies.scala
SamTheisens: Addressed in 87d3301898f6bbf86f7167a2c141cc2bbd9e8370

actors-akka/src/test/scala/reactivemongo/actors/testkit/package.scala
```diff
@@ -0,0 +1,9 @@
+package reactivemongo.actors
+
+import akka.actor.Actor
+import akka.testkit.{ TestActorRef => PekkoTestActorRef }
```
cchantep, line 4: !!

build.sbt
cchantep: all other modules are prefixed

project/Publish.scala
```diff
@@ -35,12 +35,23 @@ object Publish {
...
+  }
+
   val siteUrl = "http://reactivemongo.org"
 
   lazy val settings = Seq(
     publishMavenStyle := true,
     Test / publishArtifact := false,
     publishTo := Some(repoUrl).map(repoName at _),
+    artifactName := { (sv: ScalaVersion, module: ModuleID, artifact: Artifact) =>
+      artifact.name + modulePostfix + "-" + module.revision + "." + artifact.extension
```
cchantep, line 53:
```suggestion
      s"${artifact.name}${modulePostfix}-${module.revision}.${artifact.extension}"
```

cchantep (approved):

actors-akka/src/test/scala/reactivemongo/actors/testkit/package.scala
SamTheisens: Addressed in [Remove unnecessary import alias](https://github.com/ReactiveMongo/ReactiveMongo/pull/1245/commits/6ce14986627867731da13377e921d45995b3acf2)

build.sbt
SamTheisens: I assume you mean something like this? [Add ReactiveMongo as prefix to actor modules](https://github.com/ReactiveMongo/ReactiveMongo/pull/1245/commits/be05bbd8c6738d0631d96d5b2b166a78a88139d2)

build.sbt
```diff
@@ -1,3 +1,6 @@
+import sbt.*
```
cchantep, line 1: Not required?

project/Common.scala
cchantep: 3.3 reintroduce bug about compiler warnings

build.sbt
SamTheisens: Addressed in https://github.com/ReactiveMongo/ReactiveMongo/pull/1245/commits/6e239cbfd14ae545ced37d3297fdb5c5cd38a1d7

project/Common.scala
SamTheisens: @cchantep should I submit an issue for upgrading to scala 3.3.x?

mdedetrich: Yes, although note that Scala 3.3.1 (not yet released) is meant to fix some regressions so maybe you can test against nightly to see if its already been fixed.

SamTheisens: The problem has unfortunately not been fixed in the latest nightly. But I did manage to pin down which scala 3 change is responsible for the compilation error. This PR reproduces it https://github.com/ReactiveMongo/ReactiveMongo/pull/1251

Fristi:
Heya, what's blocking this at the moment? Is there a way I could help?

cchantep:
> Heya, what's blocking this at the moment? Is there a way I could help?

https://github.com/ReactiveMongo/ReactiveMongo/pull/1245#discussion_r1288704692

SamTheisens:
> > Heya, what's blocking this at the moment? Is there a way I could help?
> 
> [#1245 (comment)](https://github.com/ReactiveMongo/ReactiveMongo/pull/1245#discussion_r1288704692)

The compiler crash is the biggest blocker I suppose. I've submitted a crash report https://github.com/lampepfl/dotty/issues/18555.
Unfortunately I haven't been able to construct a minimal reproduction example. That would undoubtedly increase the chances of it being addressed. 
@cchantep do you perhaps have some ideas about what might be the root cause?

Fristi:
Would releasing ReactiveMongo with Pekko support and Scala 3.3 support in a seperate PR / version an idea once this is resolved? I would like to move forward with my Akka migration and I'm on a Scala 2.x code base :)

mdedetrich:
> The compiler crash is the biggest blocker I suppose. I've submitted a crash report [lampepfl/dotty#18555](https://github.com/lampepfl/dotty/issues/18555). Unfortunately I haven't been able to construct a minimal reproduction example. That would undoubtedly increase the chances of it being addressed. @cchantep do you perhaps have some ideas about what might be the root cause?

Some compiler crashes were resolved in 3.3.1, are you still having issues swith this version?

SamTheisens:
@cchantep I've found that this change https://github.com/SamTheisens/ReactiveMongo/pull/2/commits/8cac59be3f9fe8cd24c22ac93a279f029ae467a8#diff-c6447c7894adc9cfe97724d939b60a87a2aca260563123fefb254d6cf6e73097R17-R18 makes the project compile. Would that be an acceptable solution?

mdedetrich:
> @cchantep I've found that this change [SamTheisens@8cac59b#diff-c6447c7894adc9cfe97724d939b60a87a2aca260563123fefb254d6cf6e73097R17-R18](https://github.com/SamTheisens/ReactiveMongo/commit/8cac59be3f9fe8cd24c22ac93a279f029ae467a8#diff-c6447c7894adc9cfe97724d939b60a87a2aca260563123fefb254d6cf6e73097R17-R18) makes the project compile. Would that be an acceptable solution?

Nice find!

If this is not acceptable, what can be done is you can split out

```scala
type Writer[A]
type Reader[A]
```

into its own trait, and have the scala 2 implementation be

```scala
type Writer[A]
type Reader[A]
```

and scala 3 be

```scala
type Writer[A] = BSONDocumentWriter[A]
type Reader[A] = BSONDocumentReader[A]
```

that way the change is isolated just to Scala 3 artifact.

cchantep:
> > @cchantep I've found that this change [SamTheisens@8cac59b#diff-c6447c7894adc9cfe97724d939b60a87a2aca260563123fefb254d6cf6e73097R17-R18](https://github.com/SamTheisens/ReactiveMongo/commit/8cac59be3f9fe8cd24c22ac93a279f029ae467a8#diff-c6447c7894adc9cfe97724d939b60a87a2aca260563123fefb254d6cf6e73097R17-R18) makes the project compile. Would that be an acceptable solution?
> 
> Nice find!
> 
> If this is not acceptable, what can be done is you can split out
> 
> ```scala
> type Writer[A]
> type Reader[A]
> ```
> 
> into its own trait, and have the `scala-2` implementation be
> 
> ```scala
> type Writer[A]
> type Reader[A]
> ```
> 
> and `scala-3` be
> 
> ```scala
> type Writer[A] = BSONDocumentWriter[A]
> type Reader[A] = BSONDocumentReader[A]
> ```
> 
> that way the change is isolated just to Scala 3 artifact.

The API won't be changed, for temporary bug on the Scala 3 compiler.
Otherwise then we will have to manage to revert back, with version compatibility complexity.

mdedetrich:
> Otherwise then we will have to manage to revert back, with version compatibility complexity.

I don't think this change has an impact regarding MiMa because its a type parameter, so if correct you can revert it without any implications (its basically adding a default type parameter which would always be overridden anyways).

cchantep:
> > Otherwise then we will have to manage to revert back, with version compatibility complexity.
> 
> I don't think this change has an impact regarding MiMa because its a type alias, so if correct you can revert it without any implications (its basically adding a default type parameter which would always be overridden anyways).

Type member are part of the API.
So that's indeed a breaking change.

mdedetrich:
> Type member are part of the API.
So that's indeed a breaking change.

I understand they are part of the API and it is changing the API, however the term "breaking" has a very explicit technical meaning (talking about bincompat/MiMa here) and I don't think its applicable here.

Undefined type members (which is what `type Writer[A]`/`type Reader[A]` are) **have** to be reified at some point, what `type Writer[A] = BSONDocumentWriter[A]` is doing is its putting a default value on that type parameter but this shouldn't effect any code that implements this API.

Even if you were to hypothetically add this change and then remove it later, it also should not effect any code implementing this API because because you dealing with an un-implemented type member which at some point has to be evaluated to a proper type.

In fact if you check the issue that caused the change at https://github.com/lampepfl/dotty/pull/16373, at least from what I understand the reason why Scala 3 is complaining in this case is exactly because it can't find a full implementation of that type after erasure (and its not like it can scan all possible implementations of that type outside of the project).

Ill try and do some hacking on the Scala 3 compiler to see if I can avoid this check if we are dealing with undefined type members, but the reason I am bringing this up is that there is a chance that you may need to provide a default value for these undefined type members if a fix doesn't get accepted (I am mentioning this because not every change in the Scala 3 compiler is considered a regression, sometimes if the improvement is warranted then you have to deal with it).

Obviously have to see what Scala 3 team says

cchantep:
> I understand they are part of the API and it is changing the API, however the term "breaking" has a very explicit technical meaning (talking about bincompat/MiMa here) and I don't think its applicable here

Please stop arguing, that's not constructive.

It breaks the API, for an issue that being fix.

project/Common.scala
```diff
@@ -51,15 +57,15 @@ object Common extends AutoPlugin {
...
     Defaults.coreDefaultSettings ++ baseSettings ++ Compiler.settings ++ Seq(
-      scalaVersion := "2.12.18",
-      crossScalaVersions := Seq(
-        scala211,
-        scalaVersion.value,
-        scala213,
-        scala31
-      ),
+      scalaVersion := scala212,
+      crossScalaVersions := scalaVersions,
```
cchantep, line 68: Only the pekko module should exclude 3.3

SamTheisens:
The scala community is moving away from Akka. Scala 2.13 libraries can be used in Scala 3 projects and it's It's not certain when and if https://github.com/lampepfl/dotty/issues/18555 will be addressed.
Therefore I think it would be great if this PR could be merged and Scala 3 binaries added later in a follow up. As suggested here https://github.com/ReactiveMongo/ReactiveMongo/issues/1244#issuecomment-1722766242. Do you think that would be possible?

project/Common.scala
```diff
@@ -10,6 +10,11 @@ object Common extends AutoPlugin {
   override def trigger = allRequirements
   override def requires = JvmPlugin
 
+  lazy val actorModule = sys.env.get("ACTOR_MODULE") match {
```
cchantep, line 13:
```suggestion
  lazy val actorModule = sys.env.getOrElse("ACTOR_MODULE", "akka")
```

cchantep:
Thanks

SamTheisens:
> Thanks

Thanks for merging @cchantep. Out of curiosity: are you planning to make a new `1.1.0-RC12` release with these changes?

Fristi:
I would be keen to see a release with these artifacts been published, a `1.1.0-RC12` would be awesome :)

SamTheisens:
> I would be keen to see a release with these artifacts been published, a `1.1.0-RC12` would be awesome :)

It's published to [sonatype](https://oss.sonatype.org/#nexus-search;quick~reactivemongo-actors-pekko) already:
```xml
<dependency>
  <groupId>org.reactivemongo</groupId>
  <artifactId>reactivemongo-actors-pekko_2.13</artifactId>
  <version>1.1.0-RC12-SNAPSHOT</version>
</dependency>
```

PR #1299, opened 2024-04-10
Copy the channelIdHint for the next request. Fixes #1185

### LINKED ISSUE #1185: `cursor id XXX not found` when streaming data from secondary
- "org.reactivemongo" %% "reactivemongo" % "1.1.0-275c4ca-RC7-SNAPSHOT",
- "org.reactivemongo" %% "reactivemongo-akkastream" % "1.1.0-RC6",
- MongoDB 5.0.13 (and 4.4.13)
- Linux 5.19.11-arch1-1
- OpenJDK 17.0.5
- scala 3.2.

Streaming data from a secondary results in the stream closing with an error after some documents.

```scala
coll
      .find(BSONDocument(), Some(BSONDocument("_id" -> true)))
      .cursor[BSONDocument](ReadPreference.secondary)
      .documentSource(nbDocs)
      .mapConcat(d => d.int("_id").toList)
      .throttle(50, 1.second)
      .runWith(Sink.foreach(println))
```

```
Exception in thread "main" DatabaseException['cursor id 8910965325409551864 not found' (code = 43)]
```

I made a [minimal reproduction program](https://github.com/ornicar/rm-issue) with a README.

To reproduce you need a replicaset with at least one secondary.

tardigradeus:
# Pull Request Checklist

* [ ] Have you added tests for any changed functionality?

## Fixes

Fixes #1185 

## Purpose

This PR aims to fix the 1185 issue.

## Background Context

In order to keep the change minimal i decided to create a specific copy method.
Altering the signature of the `getMoreOpCmd` to give the `channelIdHint` or switching `RequestMaker` to a case class is also possible.

Feel free to change the implementation or give any feedback on my code.

About the tests, adding a secondary node to the replicaset seemed like a comprehensive change and altered the pipeline quite a bit so i did not feel confident to change this. 

The [minimal reproduction program](https://github.com/ornicar/rm-issue) written by @ornicar can validate that the issue is fixed.

## References

Are there any relevant issues / PRs / mailing lists discussions?

### DIFF
--- a/driver/src/main/scala/api/DefaultCursor.scala
+++ b/driver/src/main/scala/api/DefaultCursor.scala
@@ -585,8 +585,7 @@ private[reactivemongo] object DefaultCursor {
         logger.trace(s"Asking for the next batch of $ntr documents on cursor #${reply.cursorID}, after ${nextOffset}: ${reqMaker.op}")
 
         def req = new ExpectingResponse(
-          requestMaker =
-            reqMaker.copyWithChannelIdHint(Some(response.info.channelId)),
+          requestMaker = reqMaker.withChannelIdHint(response.info.channelId),
           pinnedNode = transaction.flatMap(_.pinnedNode)
         )
 
--- a/driver/src/main/scala/api/DefaultCursor.scala
+++ b/driver/src/main/scala/api/DefaultCursor.scala
@@ -585,7 +585,8 @@ private[reactivemongo] object DefaultCursor {
         logger.trace(s"Asking for the next batch of $ntr documents on cursor #${reply.cursorID}, after ${nextOffset}: ${reqMaker.op}")
 
         def req = new ExpectingResponse(
-          requestMaker = reqMaker,
+          requestMaker =
+            reqMaker.copyWithChannelIdHint(Some(response.info.channelId)),
           pinnedNode = transaction.flatMap(_.pinnedNode)
         )
 
--- a/driver/src/main/scala/core/protocol/RequestMaker.scala
+++ b/driver/src/main/scala/core/protocol/RequestMaker.scala
@@ -37,18 +37,15 @@ private[reactivemongo] final class RequestMaker(
     current.result()
   }
 
-  def copyWithChannelIdHint(
-      newChannelIdHint: Option[ChannelId]
-    ): RequestMaker = {
+  @inline def withChannelIdHint(channelId: ChannelId): RequestMaker =
     new RequestMaker(
       kind,
       op,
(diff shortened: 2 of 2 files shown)


driver/src/main/scala/api/DefaultCursor.scala
```diff
@@ -585,7 +585,8 @@ private[reactivemongo] object DefaultCursor {
         logger.trace(s"Asking for the next batch of $ntr documents on cursor #${reply.cursorID}, after ${nextOffset}: ${reqMaker.op}")
 
         def req = new ExpectingResponse(
-          requestMaker = reqMaker,
+          requestMaker =
+            reqMaker.copyWithChannelIdHint(Some(response.info.channelId)),
```
cchantep, line 589: Directly call the constructor rather than introducing a function used only once there

cchantep:
Hi, the change was intended, so a strict reproducer is needed to considered such revert.

tardigradeus:
> Hi, the change was intended, so a strict reproducer is needed to considered such revert.

Hi, thanks for the feedback. 
I am wondering what exactly you are calling a strict reproducer. Is this [minimal reproduction program](https://github.com/ornicar/rm-issue) not enough to assess the bug?

driver/src/main/scala/api/DefaultCursor.scala
tardigradeus: In the `RequestMaker` class there are two private fields `kind` and `parentSTE`. I cannot build the same request from the previous using constructor due to this fields. 

Would you rather create a constructor taking the previous request and the field that need to be changed?

cchantep:
I don't reproduce the issue with it and 1.1.0-RC12

tardigradeus:
Alright, i forked and changed the reproduction program in order to change the way of reproducing the issue. 

Instead of a stream I query all documents in the collection using different `ReadPreference`. And i linked the 1.1.0-RC12 version directly in the dependancies.

You'll need to have a replica set with at least 3 nodes, 1 primary and 2 secondaries. The repository can be found [here](https://github.com/tardigradeus/rm-issue)

cchantep:
Hi, even if I can consider this change, I don't reproduce the issue it's supposed to address using the provided MVR.

fitztrev:
I was able to reproduce the issue and verify the PR's fix with a modified MVR https://github.com/fitztrev/rm-issue

Updated instructions are in the readme. It uses Docker Compose to setup the mongo instances and configure the replica set.

cchantep:
Thx

PR #1372, opened 2025-03-18
GetMore OpMsg

cchantep:


cchantep:
@copilot the update of the svelte dependency also requires to update the svelte file to migrate the code

